PR #14, opened 2023-02-21
Capella zebra

osfameron:
See https://github.com/couchbasecloud/couchbase-cloud/pull/18906
and https://github.com/couchbase/docs-site/pull/659


sarahlwelton (approved):
This is such a small change. If we know that we want the version name to be "provisioned," I don't see any issue with this.

PR #15, opened 2023-02-22
NO ISSUE: Feature flag tooling compatibility

simon-dew:
Make feature flags compatible with feature flag tooling. Server-side PR.


simon-dew:
Companion to #16

simon-dew:
Adding @RichardSmedley to check that I'm not using any existing flag names incorrectly. I've listed the feature flags that I've used in the docs-devex repo at [DOC-10799](https://issues.couchbase.com/browse/DOC-10799#comment-659331).

simon-dew:
Further to discussion on Slack, I've hidden the pages that were previously embargoed. I've left the feature flag markup in place around the page-embargo attribute, so that we can remember in future why any particular page is hidden.

simon-dew:
Merging this now before my handover / training call later today.

PR #16, opened 2023-02-22
NO ISSUE: Feature flag tooling compatibility

simon-dew:
Make feature flags compatible with feature flag tooling. Cloud-side PR.


simon-dew:
Companion to #15

modules/develop/partials/escape-hatch.adoc
```diff
@@ -2,6 +2,7 @@ This
 ifdef::page-topic-type[{page-topic-type}]
 ifndef::page-topic-type[page]
 is for Couchbase Capella.
-// tag::link[]
-For Couchbase Server, see xref:server:develop:{docname}.adoc[].
-// end::link[]
\ No newline at end of file
+ifdef::flag-escape-hatch[]
+// hide-escape-hatch enables authors to override the escape hatch for individual pages
```
osfameron: ooo, cute

osfameron (approved):
This looks good!

Occurs to me that we're going to have "fun" namespacing our flags (e.g. what if someone on mobile also implements a `flag-query` etc... that's maybe a problem to consider when someone other than you starts using them... ping @RichardSmedley who's thinking about it...

simon-dew:
> Occurs to me that we're going to have "fun" namespacing our flags (e.g. what if someone on mobile also implements a `flag-query` etc... that's maybe a problem to consider when someone other than you starts using them... ping @RichardSmedley who's thinking about it...

I've listed the feature flags that I've used in the docs-devex repo at [DOC-10799](https://issues.couchbase.com/browse/DOC-10799#comment-659331).

simon-dew:
Further to discussion on Slack, I've hidden the pages that were previously embargoed. I've left the feature flag markup in place around the `page-embargo` attribute, so that we can remember in future why any particular page is hidden.

modules/develop/partials/escape-hatch.adoc
simon-dew: But unnecessary 😞 — if you [soft set](https://docs.antora.org/antora/latest/playbook/asciidoc-attributes/#soft-set) the `flag-devex-escape-hatch` flag when building the site, authors can override the escape hatch by unsetting the flag for individual pages.

simon-dew:
Merging this now before my handover / training call later today.

PR #20, opened 2023-03-16
[DOC-10862] Copy n1ql module from release/7.2 to capella

dgnicholson:
Copy n1ql module from `release/7.2` branch to `capella` branch.


simon-dew (approved):
LGTM 👍

PR #22, opened 2023-04-04
[DOC-10862] Migrate n1ql module to elixir branch

dgnicholson:
This copies the `n1ql` module from the `capella` branch to the `elixir` branch to allow for references from the serverless site.


simon-dew (approved):
Assuming this is just a straightforward copy? LGTM 👍

PR #42, opened 2023-04-25
[AV-43732] Search Developer Guide for Server 7.2

sarahlwelton:
The long awaited PR for Search. 

Please review the content on the staging site, it's much easier than trying to review all of the raw content: 

https://docs-staging.couchbase.com/server/7.2/search/search.html

PR #43, opened 2023-04-25
[AV-43732] Search Developer Guide for Capella

sarahlwelton:
The long awaited Search developer guide. 

This is the partner PR for #42. 

You can review the documentation here: 

https://docs-staging.couchbase.com/cloud/search/search.html


modules/search/pages/create-child-field.adoc
```diff
@@ -0,0 +1,120 @@
+= Create a Child Field
+:page-topic-type: guide
+:description: Create a child field with the Couchbase Capella UI to add or remove a specific field's content from a Search index.
```
bfavini, line 3: Do we need tosay "with the Couchbase Capella UI?"

sarahlwelton: Yes, so they know that they're in the Capella Developer Guide and not Server.

osfameron (approved):
Giving unblocking approval based on review of #42

PR #44, opened 2023-05-01
Search review

jon-strabala:
I am only 1/2 done.


modules/search/examples/autocomplete-search-index.jsonc
```diff
@@ -1,10 +1,8 @@
 {
     "type": "fulltext-index",
-    "name": "e-ngram-2-8",
-    "uuid": "$GENERATED_INDEX_UUID",
+    "name": "$INDEX_NAME",
     "sourceType": "gocbcore",
     "sourceName": "$BUCKET_NAME",
-    "sourceUUID": "$BUCKET_UUID",
```
sarahlwelton, line 7: Does the sourceUUID and UUID need to be deleted for every index example? I'm confused why you deleted it for this one and not the others.

modules/search/pages/create-search-indexes.adoc
```diff
@@ -16,20 +17,33 @@ You can copy the JSON payload from the UI to create your index with the REST API
...
+//* The bucket, scope and collection where you want to create the index.
+//* A xref:customize-index.adoc#type-mappings[type mapping] for the documents that you want to add to the index.
 
-After you create a Search index, you can customize it to improve search results and performance.
-For more information about how you can customize a Search index with the Capella UI, see xref:customize-index.adoc[].
+Once the index is created, the system streams data from the chosen collection(s) plus all subsequent mutations into the index builder.
+
+Search indexes can be editied and further customized or modified to enhance search results and performance. For more details on customizing a Search Index using the Capella UI, refer to xref:customize-index.adoc[Customize a Search Index with the Capella UI].
+
+[#sdks]
```
sarahlwelton: This is a much longer-term project and I would feel better having this commented out for now.

modules/search/pages/search-query-auto-complete-code.adoc
```diff
@@ -20,14 +20,26 @@ Create and configure a Search index that supports autocomplete, then configure y
...
+| "name": "$INDEX_NAME, | "name": "e-ngram-2-8",
+| "sourceName": "$BUCKET_NAME", | "sourceName": "travel-sample",
+| "$SCOPE_NAME.$COLLECTION_NAME": { | "inventory.hotel": {
+| "$FIELD_NAME": { | "name": {
+| "name": "$FIELD_NAME", | "name": "name",
+|===
+. xref:search-query-auto-complete-ui.adoc#search-index[Create an autocomplete Search index with the UI] or xref:create-search-index-rest-api.adoc[] with the file you created. 
 [start=2]
-. xref:simple-search-rest-api.adoc[Run a Search query] with 2-8 characters in the `query` property to test the Search index.
+. xref:simple-search-rest-api.adoc[Run a Search query] with 2-8 characters in the `query` property to test the Search index.  For the sample index "e-ngram-2-8," test the strings B, Be, Bea, Beau, and Beauf to locate the "Beaufort Hotel." Please note that the single character B is disregarded, as a minimum of two characters is required.
```
sarahlwelton: We don't know that they're using this against the travel sample index so I'm not recommending this.

modules/search/examples/autocomplete-search-index.jsonc
jon-strabala: In short yes if you want to create an index.

If you use the REST API the UUID specifies that this is an update to an existing index - thus if you want to create an index you must delete this value.  The self managed UI requires a delete, Eben might be actually "strippng this" in his newer UI's fo ELixir and Capella V2. 

Next sourceUUID is locked to a "bucket" so if you are trying to create a new index on the same cluster it could stay in BUT if you are moving the definition to a new cluster the sourceUUID will not match the bucket name and a create index will fail.

modules/search/pages/create-search-indexes.adoc
jon-strabala: It was "_After you create a Search index, you can customize it to improve search results and performance_" which wasn't quite correct.  I was just trying to correct the existing text. Not sure if it is a long term project, but it might be stuffed a bit deeper in the developer path.

modules/search/pages/search-query-auto-complete-code.adoc
jon-strabala: This is such a detailed item that without a concrete example on a keyspace and a field it might fail and frustrate the user. For example if the field that is to be a autocomplete is nested three levels deep the doc text just won't work.

At a minimum the current item below should change:

13. Create a Child Field with the following settings: .....

to

13. [Optional] If needed create Child Mapping until you get to the correct level in the JSON document ....
14. Create a Child Field with the following settings: .....

modules/search/pages/customize-index.adoc
```diff
@@ -13,11 +13,13 @@ You can add the following components and configure the following options for a S
 
 |[[type-identifiers]]Type Identifier a|
 
-Set a type identifier to change how the Search index finds a document's type.
+The Type Identifier offers three optional methods for filtering documents from your collections to be included in a Search index:
 
-The type identifier tells the Search Service the document field or location in the document's ID value to use as the document's type. 
+* JSON Type Field: Selects only documents that contain a specific field with a particular string value.
```
sarahlwelton: Can we please get some clarity from say, Abhi, on what exactly the type identifier does? The equivalent parameters on the JSON, you didn't edit like this, and yet now it's apparently wrong here? I want to know what exactly this does if I've apparently gotten it wrong.

modules/search/pages/set-type-identifier.adoc
```diff
@@ -16,19 +16,32 @@ For more information, see xref:create-search-index-ui.adoc[].
...
-.. To use the document's ID value, up to a given separator, as the document's type, select *Doc ID up to separator*. 
-.. To use the document's ID value with a regular expression applied as the document's type, select *Doc ID with regex*.
-. Enter the field name, separator, or regular expression in the corresponding field. 
+.. To use  *JSON type field* to select only documents that contain a specific field with a particular string value
+... Select *JSON type field*.
+... In `JSON type field` text area enter the name of the field in the document you wish to filter on.
+... Under *Type Mappings*, next to the type mapping you wish to filter, go to menu:More Options (&vellip;)[Edit]:
+... add a period `.` and then the exact `string` (with no quotes) you wish to filter on to the end of the scope.collection.
+.. To use *Doc ID up to separator* to select only documents with IDs that match a prefix up to seperator string.
+... Select *Doc ID up to separator*.
```
sarahlwelton: Can we come up with a good example, here?

jon-strabala: I recorded a brief video for you so you can see each of the three Type Identifier filters working visually as I explain what is going on ... the possible examples should be obvious.

*update* - redid the video - wasn't happy with the 1st pass

https://couchbase.zoom.us/rec/share/p2cCPJWJ2BPZgXn1UA9uy-IeUcE8Fwrhhq5tIUKtb6upzaoR9rmoqg-lEBSrrelg.D0S1K0hvZ7uFI8kC?startTime=1683141979000
Passcode: G0jNBL0%

_Note in the video I can't edit the Type Mapping more than one due to a bug I found today SURF-5559 which is being addressed in the Capella UI - so in that case I used the Self Manage editor hooked up to the same capella cluster._

sarahlwelton:
@jon-strabala , I want to get the changes in for the type identifier as this is currently published and incorrect, and has sat for long enough. Please open a new PR for any other things you want to correct with a review.

PR #53, opened 2023-06-08
AV-55075: add Tab Sync Group

osfameron:
Noticed that on https://docs-staging.couchbase.com/server/current/guides/updating-data.html the 3rd set of tabs don't synchronise, because `cbc-subdoc` is a different string than `cbc`

This commit sets them all in the same "sync group" so they match for e.g. .NET, Java, etc.

(cbc and cbc-subdoc still won't synchronize though. That doesn't seem to be customizable from https://github.com/asciidoctor/asciidoctor-tabs#tabs-sync Consider rewording both to `cbc` or `CLI` or similar if that would be desirable.)

### DIFF
--- a/modules/guides/pages/bulk-operations.adoc
+++ b/modules/guides/pages/bulk-operations.adoc
@@ -138,7 +138,7 @@ To read multiple documents in Couchbase perform a bulk `get` operation.
 
 [{tabs}]
 ====
-cbc::
+CLI::
 +
 --
 Use the `cbc cat` command to retrieve multiple documents by their IDs. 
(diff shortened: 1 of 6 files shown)


simon-dew (changes requested):
This works beautifully, and vastly improves the user experience for tabs. I'd be happy to change the `cbc-subdoc` tabs to match the `cbc` tabs to ensure that they sync correctly. But:

1. This affects more than just the updating-data page; need update all pages with `cbc` / `cbc-subdoc` tabs.
2. We need to decide on what to call the unified command line tabs. The obvious answer would be to rename `cbc-subdoc` to `cbc`. But this is inaccurate as they are two separate tools. Maybe better to standardise on `CLI` for the CLI tabs, as you suggested.
3. Just to note that the links to the SDK API documentation is broken. Should be https://docs.couchbase.com/sdk-api/couchbase-c-client/md_doc_2cbc-subdoc.html

Happy for this to be a separate PR from this ticket or a new ticket.

PR #68, opened 2023-07-07
Update datefun.adoc

GauravJayaraj:
fix comma position in arguments to weekday_millis function


simon-dew:
Thanks for the contribution! Having reviewed, I think that the text was correct as it was.

* `WEEKDAY_MILLIS(expr [, tz ])` implies that users only need to include the comma if they include the second argument, which is correct.
* `== WEEKDAY_MILLIS(expr, [ tz ])` implies that users always need to include the comma, even if the second argument is not included, which is incorrect.

Sorry that we have not included your contribution this time. We look forward to more pull requests!

PR #72, opened 2023-07-26
3rd party integrations

RichardSmedley:

PR #84, opened 2023-08-07
Partner integrations link

RichardSmedley:
AV-57-3737

### DIFF
--- a/modules/develop/pages/intro.adoc
+++ b/modules/develop/pages/intro.adoc
@@ -84,3 +84,11 @@ The Analytics Service is a parallel data management capability for Couchbase Cap
 
 * xref:analytics:introduction.adoc[Introduction to Analytics]
 endif::flag-devex-analytics[]
+
+
+== Integrations
+
+Integrations that Couchbase or partners have developed to enhance Couchbase, Capella, and partner services --
+a vital tool for many larger development projects.
+
+* xref:server:develop:integrations.adoc[]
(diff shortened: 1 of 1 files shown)


modules/develop/pages/intro.adoc
```diff
@@ -83,4 +83,12 @@ ifdef::flag-devex-analytics[]
...
\ No newline at end of file
+endif::flag-devex-analytics[]
+
+
+== Integrations
+
+Integrations that Couchbase or partners have developed to enhance Couchbase, Capella, and partner services --
+a vital tool for many larger development projects.
+
+* xref:server:develop:integrations.adoc[]
```
simon-dew, line 94: See comments on #72 . I'm really not keen on the idea of an xref in a top-level landing page on the Capella side taking users over the server side without much warning. Remember that docs-devex is specifically for information which is shared between cloud and server. If you're including the Integrations page in docs-devex on the server side, and you're linking to it from the intro page on the Capella side, you should include the Integrations page on the Capella side also.

simon-dew: (Which would mean that this link would then become

```suggestion
* xref:develop:integrations.adoc[]
```

in _this_ branch also.)

simon-dew (approved):
LGTM for now 👍  &mdash; reserving the right to argue about including integrations.adoc in the Capella branch later 😁

PR #86, opened 2023-08-15
update dotnet code sample links

tom-rosewell:
WIP - Will need to be backported to earlier server releases (and maybe capella and elixir)


RichardSmedley (approved):
:+1:

PR #95, opened 2023-11-10
AV-61582 Add Backticks Note For Travel Sample

sbirch25:
AV-61582 
Add a Backticks Note For the Travel Sample data set:
Selecting Data
Specifying a Data Source

### DIFF
--- a/modules/guides/pages/select.adoc
+++ b/modules/guides/pages/select.adoc
@@ -119,8 +119,6 @@ The set of fields returned by the query is known as the _projection_.
 NOTE: Before you can query a data source, there must be an index on that data source.
 The sample dataset already contains all the indexes you need to run the example queries in this guide.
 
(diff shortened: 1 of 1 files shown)


modules/guides/pages/select.adoc
```diff
@@ -119,6 +119,8 @@ The set of fields returned by the query is known as the _projection_.
 NOTE: Before you can query a data source, there must be an index on that data source.
 The sample dataset already contains all the indexes you need to run the example queries in this guide.
 
+NOTE: We use backticks around the travel-sample dataset because its name contains a hyphen.
```
simon-dew: I would put this in the example itself, i.e. after the query but before the result. Also I would make it just a line of text, not a note, just trying to keep visual noise and distraction to a minimum.

simon-dew (changes requested):
Just a minor comment

PR #97, opened 2023-11-15
[AV-67423] Adding information on new IP type added in 7.2

sarahlwelton:
Adding IP type to field-data-types-reference.adoc

Adding IP data searches to search-request-params.adoc

Adding example query as query-ip-range.jsonc


ElliotFrancisHunter (approved):
LGTM 👍🏼

PR #108, opened 2023-12-05
DOC-11516: Remove references to N1QL_SYSTEM_BUCKET

simon-dew:
Docs issue: DOC-11516

Draft documentation: [UPDATE STATISTICS › Optimizer Statistics](https://github.com/couchbaselabs/docs-devex/blob/7f575ab26d2581dbd2ba12ebe3dd04477ad8a999/modules/n1ql/pages/n1ql-language-reference/updatestatistics.adoc#optimizer-statistics)

Please add any comments in the **Add a comment** section below

### DIFF
--- a/modules/n1ql/pages/n1ql-language-reference/updatestatistics.adoc
+++ b/modules/n1ql/pages/n1ql-language-reference/updatestatistics.adoc
@@ -91,32 +91,25 @@ For further details, refer to xref:learn:services-and-indexes/indexes/storage-mo
 == Optimizer Statistics
 
 For each bucket, the `UPDATE STATISTICS` command stores statistics in a collection called `_query` within a scope called `_system`.
-This scope and collection are created automatically within the bucket, if they do not already exist.
+The `_system` scope and its collections, including `_query`, are created automatically when you create a bucket, or when you migrate a bucket from a previous version of Couchbase Server.
 
 === Security for Optimizer Statistics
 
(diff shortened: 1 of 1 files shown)


miaobingjie:
The system collection (_system._query) has one important difference from the previous N1QL_SYSTEM_BUCKET, it cannot be created/dropped (even with Admin privilege). The system scope and its underlying collections (there are two currently, _query and _mobile) are automatically created when a bucket is created, or in case of migration from an older version, during the migration process. You'll need to change the wording about automatically creating it if it does not exist, or Admin can drop it, or the part about it being automatically created if it does not exist.
You can also probably remove the section on memory usage. That was relevant when N1QL_SYSTEM_BUCKET is a separate bucket and automatically created. Now the system collection is within a user bucket and just shares the memory for that bucket.

simon-dew:
> The system collection (_system._query) has one important difference from the previous N1QL_SYSTEM_BUCKET, it cannot be created/dropped (even with Admin privilege). The system scope and its underlying collections (there are two currently, _query and _mobile) are automatically created when a bucket is created, or in case of migration from an older version, during the migration process. You'll need to change the wording about automatically creating it if it does not exist, or Admin can drop it

Done

> You can also probably remove the section on memory usage. That was relevant when N1QL_SYSTEM_BUCKET is a separate bucket and automatically created. Now the system collection is within a user bucket and just shares the memory for that bucket.

Done

PR #117, opened 2024-01-22
[DOC-11803] Moving Eventing module from Server to Devex for 7.6 release

julia-cb:
This PR is for moving the Eventing module from docs-server to docs-devex for the 7.6 release. There are no changes to content of docs.

Also removed module from docs-server for the 7.6 release here: https://github.com/couchbase/docs-server/pull/3432


simon-dew:
LGTM 👍 (I'm just looking at filenames, haven't checked the content)

PR #143, opened 2024-03-05
DOC-10667 Add optional GROUP AS clause for queries

lamagnifica:
Please review for completeness and accuracy

[DOC-10667](https://issues.couchbase.com/browse/DOC-10667) adds the optional GROUP AS clause, already available for SQL++ for analytics, to SQL++ for Query.

GROUP AS preserves the individual members of a group formed by a GROUP BY clause so that they can still be accessed by a subquery.

Preview site for the [updated GROUP BY topic ](https://preview.docs-test.couchbase.com/gas/server/current/n1ql/n1ql-language-reference/groupby.html)with [credentials](https://couchbasecloud.atlassian.net/wiki/spaces/DOCS/pages/1971224949/Docs+Team+demo+site+passwords).

This PR updates the GROUP BY clause description and syntax, adds GROUP AS description, syntax, and examples. I verified the provided examples using the server:7.6.1-3094 image.


dhanyagowrish (changes requested):
i.) Can we add this bit from the DOC description - "For each group, the preserved objects are stored in an array. The name of the array is the alias in the GROUP AS clause"

ii) Also - My apologies, I did not explicitly add this in the DOC description but rather pasted a link to the Analytics docs' description docs instead ( exact location - https://docs.couchbase.com/server/current/analytics/3_query.html#Group_As_clauses:~:text=The%20extra%20level%20named )

But can we please add this bit in the query docs - 

Each object in the group is wrapped in an outer object with the name of the object’s alias in the FROM clause.
To illustrate this - if the FROM clause was - `FROM airline as a, route as r`. And the GROUP AS clause is `GROUP AS g`. Each group would contain both airline and route objects. And these objects can contain field names of the same name. To avoid ambiguity, each of the original objects is wrapped in an "outer" object that gives it the name of its alias in the FROM clause.

The grouped items array ‘g’ would be:
```
[
	{
		"a": { an original airline object },
		"r": { an original route object }
	},
	{
		"a": { an original airline object },
		"r": { an original route object }
	},
	...
]
```

lamagnifica:
@dhanyagowrish for i), I reworded sentences that originated from the analytics text. See [this commit](https://github.com/couchbaselabs/docs-devex/pull/143/commits/39419ce2f79b9d57ccab2c328ffe72a4d003db3b).

For ii) I reworded your text so that it made sense to me (so it might well be completely incorrect now).  I wasn't sure where you wanted me to add this, so it is with the description of GROUP AS rather than after the two more straightforward examples at the end of the topic. 
I also tweaked one of the original examples to include a FROM clause alias and show the wrapping in the results. See [this commit](https://github.com/couchbaselabs/docs-devex/pull/143/commits/da583833c9264498b0a8f7e01bba968950349141).

modules/n1ql/pages/n1ql-language-reference/groupby.adoc
```diff
@@ -49,6 +50,51 @@ For details, see xref:n1ql-language-reference/from.adoc#section_ax5_2nx_1db[AS C
...
+A query can include a `GROUP AS` clause only if it has a `GROUP BY` clause. 
+The effect of a `GROUP BY` clause is to hide the original objects in each group, exposing only the grouping expressions and special aggregation functions on the non-grouping fields. 
+The purpose of the `GROUP AS` clause is to make the original objects in the group visible to subsequent clauses. 
+As a result, the query can generate output data both for the group as a whole and for the individual objects inside the group.
+
+For each group, `GROUP AS` preserves all of the objects in the group, as they were before grouping, in an array. 
+The name of the array is the alias in the GROUP AS clause. 
+You can then use the array name in the `FROM` clause of a subquery to process and return the individual objects in the group.
+
+If the `FROM` clause of the query includes an alias or aliases, the `GROUP AS` array wraps each object in an outer object identified by its `FROM` alias.
```
dhanyagowrish: Please change this line "If the `FROM` clause of the query includes an alias or aliases, the `GROUP AS` array wraps each object in an outer object identified by its `FROM` alias. "

... to something along the lines of - 

In the GROUP AS array, each object in the group is wrapped in an outer object that gives it the name of the alias if specified or the implicit alias of its corresponding from-term in the FROM clause.

[ The reason I ask for this change is because the FROM clause need not explicitly have an alias mentioned.  If no alias is set example if simply like - SELECT * FROM airline; the alias of the "airline" from-term is simply "airline". This is the implicit alias ]

dhanyagowrish (changes requested):

modules/n1ql/pages/n1ql-language-reference/groupby.adoc
```diff
@@ -12,7 +12,7 @@ include::partial$n1ql-language-reference/horizontal-style.adoc[]
 == Purpose
 
 Use the GROUP BY clause to arrange aggregate values into groups of one or more fields.
-This `GROUP BY` clause follows the `WHERE` clause and precedes the optional `LETTING`, `HAVING`, and `ORDER BY` clauses.
+This `GROUP BY` clause follows the `WHERE` clause and precedes the optional `GROUP AS`, `LETTING`, `HAVING`, and `ORDER BY` clauses.
```
simon-dew: Only just noticed this. I don't think it's right to say that the GROUP BY clause _precedes_ GROUP AS, LETTING, HAVING, or ORDER BY, in as much as these are part of the GROUP BY clause, whereas ORDER BY is on the same level. I would rephrase this to say:

```suggestion
This `GROUP BY` clause follows the `WHERE` clause and may contain the optional `GROUP AS`, `LETTING`, and `HAVING` clauses.
```

simon-dew (changes requested):
Only one minor change from me.

modules/n1ql/pages/n1ql-language-reference/groupby.adoc
lamagnifica: ah, I see it now, thank you

modules/n1ql/pages/n1ql-language-reference/groupby.adoc
```diff
@@ -49,6 +50,51 @@ For details, see xref:n1ql-language-reference/from.adoc#section_ax5_2nx_1db[AS C
...
+
+For each group, `GROUP AS` preserves all of the objects in the group, as they were before grouping, in an array. 
+The name of the array is the alias in the GROUP AS clause. 
+You can then use the array name in the `FROM` clause of a subquery to process and return the individual objects in the group.
+
+In the `GROUP AS` array, each object is wrapped in an outer object that gives it the name of the alias specified in the `FROM` clause, or its implicit alias if no alias was specified.
+Identifying the objects with their aliases in this way helps avoid ambiguity.
+
+For example, a query has a `FROM` clause of `FROM airline as a, route as r`, and a `GROUP AS` clause of `GROUP AS g`. 
+The `GROUP AS` clause preserves the airline and route objects together in the same array.
```
dhanyagowrish: One minor change request - could we please change "The `GROUP AS` clause preserves the airline and route objects together in the same array. " 

to

"Each item in the group would contain both an `airline` object and a `route` object."

[ the reason I ask for this change is because it is important for users to understand that each _group item_ will have both objects. And hence, requiring each object to be identified separately with its alias ]

Everything else looks great to me!

dhanyagowrish (changes requested):

modules/n1ql/pages/n1ql-language-reference/groupby.adoc
lamagnifica: got it. 
I am trying very hard not to use the word "group" to describe the JSON syntax that results from GROUP BY or GROUP AS since I think that overloads the word. 
How's this @dhanyagowrish : 

The result of the `GROUP AS` clause is an array of objects, each of which contains both an airline object and a route object.

modules/n1ql/pages/n1ql-language-reference/groupby.adoc
```diff
@@ -49,6 +50,51 @@ For details, see xref:n1ql-language-reference/from.adoc#section_ax5_2nx_1db[AS C
...
+The name of the array is the alias in the GROUP AS clause. 
+You can then use the array name in the `FROM` clause of a subquery to process and return the individual objects in the group.
+
+In the `GROUP AS` array, each object is wrapped in an outer object that gives it the name of the alias specified in the `FROM` clause, or its implicit alias if no alias was specified.
+Identifying the objects with their aliases in this way helps avoid ambiguity.
+
+For example, a query has a `FROM` clause of `FROM airline as a, route as r`, and a `GROUP AS` clause of `GROUP AS g`. 
+The `GROUP AS` clause preserves the airline and route objects together in the same array. 
+These objects might contain field names that are the same, such as `"id"`. 
+However, the resulting array `g` identifies each of the object separately with its alias, as follows:
```
lamagnifica: I'll delete "However, " here as unnecessary, fix up to:

The resulting array `g` identifies each of the objects it contains separately with its alias, as follows:

modules/n1ql/pages/n1ql-language-reference/groupby.adoc
dhanyagowrish: That sounds great.

simon-dew (approved):

simon-dew:
I'm happy with this when you've made all of @dhanyagowrish's changes.

dhanyagowrish (approved):

PR #146, opened 2024-03-08
Changes for Capella vector search

jon-strabala:
1. Made things consistent with the sample data (via pendo) in which there is no colorvect_dot
2. Made Capella friendly HTTPS (no HTTP in Capella) and used proper port number 18094
3. Tried to organize the flow of the intro page better and split out similarity and semantic search
4. Fixed all examples to actually work both in the Capella UI and on the command line
5. Wrapped $EVAR as ${EVAR} to avoid issues with syntax like  $EVAR1:$EVAR2 where the boundaries of a variable name are a problem


jon-strabala:
@sarahlwelton, I only worked on the vector stuff in Capella today feel free to fix grammer ro other messups.

PR #158, opened 2024-03-15
Need "var' to make legal

jon-strabala:
Julia minor bug will prevent the examples form working


julia-cb (approved):

julia-cb:
Thanks Jon! Just merged the fix.

PR #171, opened 2024-03-21
[AV-75382] Removing all mentions of REST API from Capella doc + Resyncing Integrations Page

sarahlwelton:
Turns out that we don't want to be advertising the REST API for Capella at all. Needed to hurriedly create this PR to remove all mention of it from the doc set. 

Also fixing the integrations page again so it stays in sync with Server.


bfavini (approved):
LGTM

PR #172, opened 2024-03-21
AV-75388: Correct xref to cloud doc

lamagnifica:
per our Slack conversation @sarahlwelton , I have corrected the cross reference link to the Capella doc file


sarahlwelton (approved):
If you've tested on a local build and this resolves, let's get it in :)

PR #173, opened 2024-03-28
[AV-75839] Add Page Aliases to Support FTS File Removal on Capella Docs

sarahlwelton:
Adding some aliases on the Capella side for deleted content from old FTS docs.

There is a related PR on couchbase-cloud for removing those files - this just makes sure that the links to those files don't just end up at a 404.

### DIFF
--- a/modules/search/pages/create-search-indexes.adoc
+++ b/modules/search/pages/create-search-indexes.adoc
@@ -2,6 +2,7 @@
 :page-topic-type: concept 
 :page-ui-name: {ui-name}
 :page-product-name: {product-name}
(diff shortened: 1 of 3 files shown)


ElliotFrancisHunter (approved):
Looks good! Only confusion I have is on the other PR you have a perform-full-text-searches.adoc AND a .perform-full-text-searches.adoc but you only have an alias for the former here? Is the dot one a dupe or something?

sarahlwelton:
> Looks good! Only confusion I have is on the other PR you have a perform-full-text-searches.adoc AND a .perform-full-text-searches.adoc but you only have an alias for the former here? Is the dot one a dupe or something?

The . file is a dupe that also wouldn't be included in the build anyway, so it's not necessary to alias.

PR #179, opened 2024-04-16
[DOC-12098] Added back ticks to create index example in WITH RECURSIVE page

julia-cb:
Small fix to add back ticks to "travel-sample" in the create index example 4 on the [WITH RECURSIVE page](https://docs.couchbase.com/server/current/n1ql/n1ql-language-reference/with-recursive.html).

Fix version on the doc ticket is 7.6.2 but since this is a fix and the code example doesn't run without the back ticks, should we just merge it right now?

Jira ticket: [DOC-12098](https://issues.couchbase.com/browse/DOC-12098)


RichardSmedley (approved):
LGTM  :+1:

simon-dew (approved):
LGTM 👍

PR #197, opened 2024-04-26
[AV-77531] Remove all mentions of non-existent sample file from Vector Search Docs

sarahlwelton:
What the title says. 

They suddenly removed the sample file that the docs were relying on, so now I had to make corrections to remove all mentions of that file. 

Fixing broken links for other non-existent docs and making some QoL changes based on other customer feedback at the same time.


ElliotFrancisHunter (approved):
LGTM 👍🏼

PR #198, opened 2024-04-26
[DOC-12154] Remove all mentions of non-existent sample file from Vector Search Docs

sarahlwelton:
Equivalent PR for server doc from https://github.com/couchbaselabs/docs-devex/pull/197 

What the title says.

They suddenly removed the sample file that the docs were relying on, so now I had to make corrections to remove all mentions of that file.

Making some QoL changes based on other customer feedback at the same time.


modules/vector-search/pages/run-vector-search-ui.adoc
```diff
@@ -1,4 +1,4 @@
-= Run A Vector Search with the Web Console 
+= Run A Vector Search with the {page-ui-name}
```
ElliotFrancisHunter, line 1: Does this work when the attribute is declared below the line?

ElliotFrancisHunter (approved):
Looks good on my end! Happy to approve as I'm assuming that title works as expected 👍🏼

modules/vector-search/pages/run-vector-search-ui.adoc
sarahlwelton: yup - you can see it's already set on the Capella side and works just fine.

PR #203, opened 2024-04-30
[AV-75759] Vector Search Specific Examples

sarahlwelton:
Whole goal of this PR is to add some more specific examples to the vector search documentation after feedback came in from Sriram. 

I added an example of a hybrid query, and added a direct "copy/paste" example Search index that people can hopefully use to just get going right away with the examples.

PR #204, opened 2024-04-30
[DOC-12168] Copy Changes from AV-75759

sarahlwelton:
Whole goal of this PR is to add some more specific examples to the vector search documentation after feedback came in from Sriram - this is the sister PR to https://github.com/couchbaselabs/docs-devex/pull/203

I added an example of a hybrid query, and added a direct "copy/paste" example Search index that people can hopefully use to just get going right away with the examples on the UI side. 

This side has the REST API, so some additional changes were necessary.

----

Adding a full sample payload that people can import for the UI topic

Adding an example section to create-vector-search-index-ui.adoc

Adding additional hybrid query example section to run-vector-search-ui.adoc

Adding recommendation to use sample vector search index in run-vector-search-sdk.adoc.

Adding recommendation to use sample vector search index in run-vector-search-rest-api.adoc.

Adding note that indexes are equivalent in create-vector-search-index-rest-api.adoc

### DIFF
--- a/modules/vector-search/pages/create-vector-search-index-ui.adoc
+++ b/modules/vector-search/pages/create-vector-search-index-ui.adoc
@@ -95,6 +95,7 @@ This Vector Search index has a type mapping for a `color.rgb` collection and inc
 
 * The *brightness number* field, which is included in search results and supports sorting and faceting.
 * The *color string* field, which is included in search results, supports highlighting, phrase matching, and sorting and faceting.
+* The *description string* field, which is included in search results, supports highlighting, phrase matching, and sorting and faceting.
 * The *colorvect_l2 [ number ]* field, which has a Dimension of `3` and uses the *l2_norm* Similarity Metric.
 * The *embedding_vector_dot [ number ]* field, which has a dimension of `1536` and uses the *dot_product* Similarity Metric. 
 
--- a/modules/vector-search/pages/create-vector-search-index-ui.adoc
+++ b/modules/vector-search/pages/create-vector-search-index-ui.adoc
@@ -1,7 +1,8 @@
-= Create a Vector Search Index with the Web Console
+= Create a Vector Search Index with the {page-ui-name}
 :page-topic-type: guide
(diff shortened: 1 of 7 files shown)


modules/vector-search/pages/create-vector-search-index-ui.adoc
```diff
@@ -78,13 +79,30 @@ For more information, see xref:search:child-field-options-reference.adoc[].
...
+----
+include::example$sample-vector-search-index-payload.json[]
+----
+
+NOTE: Make sure you imported the sample dataset with the recommended settings. 
+
+This Vector Search index has a type mapping for a `color.rgb` collection and includes the following fields: 
+
+* The *brightness number* field, which is included in search results and supports sorting and faceting.
+* The *color string* field, which is included in search results, supports highlighting, phrase matching, and sorting and faceting.
```
julia-cb: Is there no "description string" field here?

modules/vector-search/pages/run-vector-search-ui.adoc
```diff
@@ -57,6 +57,24 @@ It only returns the `k` number of similar vectors.
 The Search Service combines the Vector search results from a `knn` object with the traditional `query` object by using an `OR` function. 
 If the same documents match the `knn` and `query` objects, the Search Service ranks those documents higher in search results. 
 
+The document for the color `navy` should be the first result, followed by a similar color. 
+
+[#hybrid]
+=== Example: Run a Simple Hybrid Search Query
```
julia-cb:
```suggestion
=== Example: Running a Simple Hybrid Search Query 
```

julia-cb (commented):
Just a small question about one of the fields! Otherwise, looks great 😄

modules/vector-search/pages/create-vector-search-index-ui.adoc
sarahlwelton: Ah, good catch. Thank you, it should be here.

PR #233, opened 2024-06-05
DOC-12086: Add shell to Capella KV guides

simon-dew:
Jira issue: DOC-12086

Adds Couchbase Shell instructions and examples to the Capella KV guides, filling a big gap in the Capella devex docs.

Doc previews:
* [Work with Documents](https://preview.docs-test.couchbase.com/DOC-12086/cloud/guides/kv-operations.html)
  - [Create Documents](https://preview.docs-test.couchbase.com/DOC-12086/cloud/guides/creating-data.html)
  - [Read Documents](https://preview.docs-test.couchbase.com/DOC-12086/cloud/guides/reading-data.html)
  - [Update Documents](https://preview.docs-test.couchbase.com/DOC-12086/cloud/guides/updating-data.html)
  - [Delete Documents](https://preview.docs-test.couchbase.com/DOC-12086/cloud/guides/deleting-data.html)
  - [Work with Documents in Bulk](https://preview.docs-test.couchbase.com/DOC-12086/cloud/guides/bulk-operations.html)


simon-dew:
@osfameron is there anything that can be done about the wonky fixed-width characters in the cbsh output? For example at [Work with Documents in Bulk › Creating Multiple Documents](https://preview.docs-test.couchbase.com/DOC-12086/cloud/guides/bulk-operations.html#creating-multiple-documents)

simon-dew:
FIXME:

- [x] In Work with Documents, **Key-Value (KV) or Data Service** should link to [Manage Your Data](https://preview.docs-test.couchbase.com/DOC-12086/cloud/clusters/data-service/data-service.html)
- [x] shebang in the .nu files [Shebangs (#!)](https://www.nushell.sh/book/scripts.html#shebangs)

osfameron:
> @osfameron is there anything that can be done about the wonky fixed-width characters in the cbsh output? For example at [Work with Documents in Bulk › Creating Multiple Documents](https://preview.docs-test.couchbase.com/DOC-12086/cloud/guides/bulk-operations.html#creating-multiple-documents)

@simon-dew I think you were right in our last chat that it's the box-drawing characters that are slightly too wide.  Not sure what can be done, but found this discussion of the exact same problem https://stackoverflow.com/questions/70797173/monospace-font-characters-are-not-fixed-width

(All of which is interesting, but I'm not sure how much of a priority it is in our case?)

simon-dew:
@osfameron that was an interesting post, thanks. Inspecting our HTML reveals that our fixed-width font for alphanumeric characters and punctuation is Source Code Pro, but the box drawing characters are Menlo. So they're not even the same font! Looking at the [Google font tester for Source Code Pro](https://fonts.google.com/specimen/Source+Code+Pro/tester), it seems the box drawing characters aren't even in the font.

Setting the font for code blocks by hand to `mono` or `Menlo` fixes the character width problem. I'm not sure how much freedom we have to set our own fonts for the documentation site though?

I don't know how much of a priority this is either, it depends how irritating we think the wonky output is.

osfameron:
I think it's just a CSS change, *but* we would probably have to:

* test it against a range of code examples to make sure we don't break something else
* OK the font with web team against Couchbase style identity

modules/guides/pages/updating-data.adoc
```diff
@@ -26,62 +25,46 @@ include::partial$data-warning.adoc[]
...
-[{tabs}]
+[tabs]
 ====
-ifdef::flag-devex-command-line[]
-cbc::
+cbsh::
 +
 --
-. Create a JSON document containing the updated data.
+. If necessary, use `cb-env` to set the bucket, scope, and collection in which to store the document.
```
dgnicholson: How would I know it's necessary? Maybe expand it a bit?

> If you have not already done so, use `cb-env` to set the bucket, scope, and collection in which to store the document.

(I don't have the full context, so it might be obvious and not worth expanding)

modules/guides/pages/updating-data.adoc
```diff
@@ -162,74 +145,40 @@ For further details, refer to {python-api-url}#collection-object[Collection^].
...
-[{tabs}]
+[tabs]
 ====
-ifdef::flag-devex-command-line[]
-cbc::
+cbsh::
 +
 --
-. Update a JSON document with some new data.
+. If necessary, use `cb-env` to set the bucket, scope, and collection in which to store the document.
```
dgnicholson: Same comment as above: https://github.com/couchbaselabs/docs-devex/pull/233/files#diff-16cc696a661eb20075cebf86b6288f02c35fbd436274b398a5b160c96b0e33e6R33

modules/guides/pages/deleting-data.adoc
```diff
@@ -22,37 +21,39 @@ include::partial$data-warning.adoc[]
...
-[{tabs}]
+[tabs]
 ====
-ifdef::flag-devex-command-line[]
-cbc::
+cbsh::
 +
 --
-Use the `cbc rm` command to delete a document from the database.
+Use the `doc remove` command to delete a document from the database.
```
dgnicholson: Do we need to add mention of using `cb-env` to set the bucket, scope, and collection? Mostly just wondering why only the insert operations specifically call out `cb-env`.

dgnicholson (approved):
A few questions, but not blockers. 👍

modules/guides/pages/updating-data.adoc
Westwooo: This may be worth changing. A bucket will always need to be set when performing any `doc`command. After that setting the scope/collection is optional, since if they aren't set then they are set to _default, this can be seen as follows:

```
👤 Westwooo 🏠 local
> cb-env bucket travel-sample
👤 Westwooo 🏠 local in 🗄 travel-sample._default._default
```

simon-dew: Done

modules/guides/pages/updating-data.adoc
simon-dew: Done

modules/guides/pages/deleting-data.adoc
simon-dew: Updated

simon-dew:
Going to merge this now as it's been looked at by at least one Couchbase technical writer and one Couchbase Shell developer. We can make further improvements when Couchbase Shell comes out of developer preview.

PR #235, opened 2024-06-06
NO ISSUE: Fix broken link in tutorial

simon-dew:
Fixes link to the Java full installation page.

Also moves the tip to use SDKMan before the link to the Java full installation page, because SDKMan is the recommended method.


RichardSmedley (approved):
:+1:

PR #245, opened 2024-07-26
DOC-12087: Add shell to Capella Query guides

simon-dew:
Jira issue: DOC-12087

This PR extensively reworks the cbq guide for use with Capella, and removes the `flag-devex-tools` feature flag so that cbq information becomes available throughout the Capella devex docs. It also makes several driveby fixes for links and terminology.

Note that the links on the [Command Line Tools](https://docs.couchbase.com/cloud/reference/command-line-tools.html) page will be updated as part of AV-80110.

Docs preview: [cbq: The Command Line Shell for SQL+⁠+](https://preview.docs-test.couchbase.com/DOC-12087/cloud/n1ql/n1ql-intro/cbq.html)


bfavini (approved):
Looks good. For the future....we may need to think about whether we need to be more specific when we say Couchbase Capella UI...not sure, but we may need to distinguish between Operational and Columnar.

simon-dew:
No further review comments received so will merge as is.

PR #254, opened 2024-08-09
[DOC-11988-capella]: RBAC role change impact to eventing function

RayOffiah:
Added cautionary note concerning changing permissions that are attached to a deployed function.
Note that the caution is wrapped in tags so that the same caution can be used in the Server docs.

### DIFF
--- a/modules/eventing/pages/eventing-rbac.adoc
+++ b/modules/eventing/pages/eventing-rbac.adoc
@@ -27,9 +27,20 @@ For more information, see xref:learn:security/authorization-overview.adoc[Author
 A bucket.scope combination is used for identifying functions belonging to the same group.
 
 Only the "Eventing Full Admin" role and also the "Full Admin" role can set the bucket.scope to  *+*+.+*+*; all other Eventing non-privileged users need to define a *Function Scope* for their Eventing functions that references an existing resource of bucket.scope. 
-This provides role based isolation of Eventing functions between non-privileged users
+This provides role-based isolation of Eventing functions between non-privileged users
 
-Typically you should set Function Scope to the bucket.scope that holds the collection that is the source of your mutations to your Eventing Function.  This best practice ensures that you _do not_  inadvertently cause an Eventing Function to undeploy by removing a *Function Scope* pointing to a resource that is not required for the function to run.
+// tag::rbac-change-warning[]
+[CAUTION]
+====
+Changing the access role (i.e.,
+by revoking write permissions) could impact deployed eventing functions that have been assigned to this role. +
+_This may result in the function being undeployed._ +
+In this case, redeploy the function with the correctly assigned role to allow access.
+====
+// end::rbac-change-warning[]
+
+Typically, you should set Function Scope to the `bucket.scope` that holds the collection that is the source of the mutations to your Eventing Function. +
+This best practice ensures that you _do not_  inadvertently cause an Eventing Function to undeploy by removing a *Function Scope* pointing to a resource that is not required for the function to run.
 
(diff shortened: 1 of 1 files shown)


modules/eventing/pages/eventing-rbac.adoc
```diff
@@ -29,7 +29,18 @@ A bucket.scope combination is used for identifying functions belonging to the sa
 Only the "Eventing Full Admin" role and also the "Full Admin" role can set the bucket.scope to  *+*+.+*+*; all other Eventing non-privileged users need to define a *Function Scope* for their Eventing functions that references an existing resource of bucket.scope. 
 This provides role based isolation of Eventing functions between non-privileged users
 
-Typically you should set Function Scope to the bucket.scope that holds the collection that is the source of your mutations to your Eventing Function.  This best practice ensures that you _do not_  inadvertently cause an Eventing Function to undeploy by removing a *Function Scope* pointing to a resource that is not required for the function to run.
+// tag::rbac-change-warning[]
```
osfameron, line 32: Does this work? I thought the file needed to be set `:page-partial:` to get picked up from the change in the other PR?

modules/eventing/pages/eventing-rbac.adoc
```diff
@@ -29,7 +29,18 @@ A bucket.scope combination is used for identifying functions belonging to the sa
...
+[CAUTION]
+====
+Changing the access role (i.e., by revoking write permissions) could impact deployed eventing functions that have been assigned to this role. It is possible that the function may be undeployed; in this case, redeploy the function with a correctly assigned role to allow access.
```
osfameron: this could be ventilated

modules/eventing/pages/eventing-rbac.adoc
```diff
@@ -29,7 +29,18 @@ A bucket.scope combination is used for identifying functions belonging to the sa
...
+====
+// end::rbac-change-warning[]
+
+Typically, you should set Function Scope to the bucket.scope
```
osfameron: this section here is oddly ventilated (e.g. line breaks not just at full stop, or even commas, but at logical chunks)?
I don't hate it, but it's maybe a little arbitrary, so more susceptible to weird diffs if next writer disagrees with the chunking.

osfameron (commented):
Looks good, lemme get update on the question about that tag before I approve though

modules/eventing/pages/eventing-rbac.adoc
RayOffiah: I think it should, but I may be wrong. Just checking again.

RayOffiah: Okay, got it working in the preview, by inserting an underscore in front of the @cloud … well, I say 'I', but IntelliJ corrected it by adding the underscore, so I'm going to have to ask the Zulip folk what the underscore does (at a guess, it's for components without a version id).

Thinking about it though; it's probably better to just put the note in a partial, as you suggested. I'll fix it on Monday

modules/eventing/pages/eventing-rbac.adoc
RayOffiah: Yeah, see what you mean. I'll take another look.

modules/eventing/pages/eventing-rbac.adoc
RayOffiah: Yup. Dan answered my question: the underscore denotes a component without a version id (the Capella component doesn't have one). 

I'm going to leave it as it is, if that's okay. 👍🏾

RayOffiah: One other small point: the reason I've put the tagged section in the Capella branch is because it is unversioned, so I figure it's best to put it somewhere that won't have multiple versions.

osfameron (approved):

PR #257, opened 2024-08-15
[AV-77245] Vector Search Sample Data from UI

sarahlwelton:
Updating partial for prerequisites TIP to point to importing sample data from the UI, rather than S3 bucket link.


Rebecca-Martinez007 (approved):
LGTM! :)

PR #258, opened 2024-08-16
Revert "[DOC-11988]: RBAC role change impact to eventing function"

RayOffiah:
Reverts couchbaselabs/docs-devex#256


RayOffiah:
Reverting. Needs another look.

PR #259, opened 2024-08-16
Revert "[DOC-11988-capella]: RBAC role change impact to eventing function"

RayOffiah:
Reverts couchbaselabs/docs-devex#254


RayOffiah:
Reverting. Seems to cause problems with menu.

PR #262, opened 2024-08-19
[DOC-11988-capella] capella rbac role change impact to eventing function

RayOffiah:
Third time's the charm!

We've put the warning inside its own **_partial$_** file, as doing otherwise made the resulting build somewhat unpredictable.

### DIFF
--- a/modules/eventing/partials/rbac-warning.adoc
+++ b/modules/eventing/partials/rbac-warning.adoc
@@ -1,0 +1,7 @@
+[CAUTION]
+====
+Changing the access role (i.e.,
+by revoking write permissions) could impact deployed eventing functions that have been assigned to this role. +
+_This may result in the function being undeployed._ +
+In this case, redeploy the function with the correctly assigned role to allow access.
+====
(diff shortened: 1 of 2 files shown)


modules/eventing/partials/rbac-warning.adoc
```diff
@@ -0,0 +1,7 @@
+[CAUTION]
+====
+Changing the access role (i.e.,
+by revoking write permissions) could impact deployed eventing functions that have been assigned to this role. +
```
osfameron, line 4: Not important, but do you need the trailing `+`?

(Just asking 'cos I never understand where you have to use this... one of the weirder asciidoc features...)

osfameron (approved):

modules/eventing/partials/rbac-warning.adoc
RayOffiah: In this case, yes. 

I wanted the next line rendered separately, so I need the `+` to stop it appearing  on the same line, without a blank line between them.

PR #271, opened 2024-09-10
minor fixes to bucket and index names in sample search index and alias JSON

BKaneAtWork:
minor fixes to bucket and index names in sample search index and alias JSON so that these work with the corresponding sample data sets


sarahlwelton (approved):
These samples got worked on before we got the sample data sorted - thanks for catching these.

PR #274, opened 2024-10-03
[quickfix]: strange overlapping  text on diagram

RayOffiah:
Weird artifact on the architecture diagram that was causing one the text on one of the labels to overlap.


osfameron (approved):
I don't see the overlap issue, but no objection to a whitespace fixup, and yay if that fixes some problem 🥳

PR #282, opened 2024-10-28
[DOC-12252] Add tip with links to other SDK examples

sarahlwelton:
Quick fix to add a tip that shows the other examples Richard wrote up for Vector Search for the other SDK languages. 

It'll at least stop further complaints, even if it's not a perfect solution. 

This is the Capella version of the fix.

### DIFF
--- a/modules/vector-search/pages/run-vector-search-sdk.adoc
+++ b/modules/vector-search/pages/run-vector-search-sdk.adoc
@@ -11,6 +11,21 @@
 
 For more information about how the Search Service scores documents in search results, see xref:search:run-searches.adoc#scoring[Scoring for Search Queries].
 
+[TIP]
+====
+Not all available Couchbase SDK languages are covered by the examples on this page.
+
+For additional Vector Search examples, see: 
+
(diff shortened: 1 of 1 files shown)


modules/vector-search/pages/run-vector-search-sdk.adoc
```diff
@@ -11,6 +11,21 @@
 
 For more information about how the Search Service scores documents in search results, see xref:search:run-searches.adoc#scoring[Scoring for Search Queries].
 
+[TIP]
+====
+Not all available Couchbase SDK languages are covered by the examples on this page.
+
+For additional Vector Search examples, see: 
+
+* xref:dotnet-sdk:howtos:full-text-searching-with-sdk.adoc#vector-search[.NET SDK]
```
RichardSmedley: LGTM - the links are fine.

But would it be better if all of the list rendered on a single line, to avoid this admonition dominating the top of the page?
Maybe a single line of pipe-separated link text,
or a more semantically correct approach of styling your bulleted list?

RichardSmedley (approved):
(It's just a thought - not blocking merging)

PR #283, opened 2024-10-28
[DOC-12252] Add tip with links to other SDK examples

sarahlwelton:
Quick fix to add a tip that shows the other examples Richard wrote up for Vector Search for the other SDK languages.

It'll at least stop further complaints, even if it's not a perfect solution.

This is the Server version of the fix.


RichardSmedley (approved):
LGTM - but see comment on https://github.com/couchbaselabs/docs-devex/pull/282

PR #290, opened 2024-11-13
[DOC-12608] Add Cosine Similarity Metric

sarahlwelton:
Adding another entry to the similarity metric setting for cosine similarity.

Only one change needed as it's in a partial already.

### DIFF
--- a/modules/search/partials/vector-search-field-descriptions.adoc
+++ b/modules/search/partials/vector-search-field-descriptions.adoc
@@ -31,6 +31,15 @@ Smaller euclidean distances mean that the values of each coordinate in the vecto
 +
 It's best to use *l2_norm* similarity when your embeddings contain information about the count or measure of specific things, and your embedding model uses the same similarity metric. 
 
+* *cosine*: Calculated by adding the result of multiplying a vector's components, or the product of the magnitudes of the vectors and the cosine of the angle between them.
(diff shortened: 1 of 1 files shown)


modules/search/partials/vector-search-field-descriptions.adoc
```diff
@@ -31,6 +31,15 @@ Smaller euclidean distances mean that the values of each coordinate in the vecto
 +
 It's best to use *l2_norm* similarity when your embeddings contain information about the count or measure of specific things, and your embedding model uses the same similarity metric. 
 
+* *cosine*: Calculated by adding the result of multiplying a vector's components, or the product of the magnitudes of the vectors and the cosine of the angle between them.
+This metric is not affected by the size of the vectors being measured.
++
+Use *cosine* similarity to get the best results with an embedding model that uses cosine similarity.
+Cosine similarity works well for semantic search, document classification, and recommendation systems.
++
+The Search Service will normalize any vectors before indexing when using cosine similarity.
```
abhinavdangeti: Worth mentioning that - we will also normalize the query vector prior to executing the nearest neighbor search if the field it is associated with is using the `cosine` similarity distance metric.

Rebecca-Martinez007 (approved):
LGTM!

PR #293, opened 2024-11-18
[DOC-12051] Clarifying instructions for creating a basic search index.

sarahlwelton:
Clearing up old doc bugs, just adding a clarification on the index creation process. Tried to go too simple in the original.


Rebecca-Martinez007 (approved):
LGTM!

PR #298, opened 2024-11-19
[DOC-10997] fts-pagination.adoc Floating Page Fix

sarahlwelton:
Folding in information from fts-pagination to prepare for deleting the page from builds.

See https://github.com/couchbase/docs-server/pull/3730


osfameron (approved):
The "floating page" fix is just the `:page-aliases:` change above?

There's a lot of driveby changes if so, I can approve to unblock, but just want to check that you intended to commit these?

PR #302, opened 2024-12-02
[DOC-12432] Document Usage of {"ctl": { "validate" : true }, } for Search Queries

simon-dew:
Docs issue: DOC-12432

Updates made by @sarahlwelton . I force-pushed to resolve a merge conflict with `prerelease/7.6.4`. 

Looks like this is an improvement to the documentation (for clarification), so I think no release notes needed for 7.6.4?


sarahlwelton (approved):
Don't see any issues. Thanks!

simon-dew:
Updates after review: made it clear that the `validate` property was introduced in 7.6.4

PR #311, opened 2025-01-15
DOC-12699: Update views deprecation notice

hyunjuV:
Update views deprecation notice to include additional info -- "Instead of views, use indexes and queries using the Index Service (GSI) and the Query Service (SQL++)".


shivaniguptasf (approved):
Looks good

PR #312, opened 2025-01-18
[DOC-12901]: erroneous vector search text

RayOffiah:
Refactor pre-filtering description in vector search guide.
The pull request simplifies and clarifies the explanation of pre-filtering benefits in the vector search guide.

### DIFF
--- a/modules/vector-search/pages/pre-filtering-vector-search.adoc
+++ b/modules/vector-search/pages/pre-filtering-vector-search.adoc
@@ -8,13 +8,10 @@
 
 == About Pre-filtering
 
-Using pre-filtering as part of your vector search offers two key advantages:
+The Search Service supports pre-filtering on Vector Search queries. 
+Pre-filtering allows you to execute vector searches over a subset of the vector index, 
+via the means of a filter request that qualifies the subset.
 
-. *Enhanced precision and relevance:*
-Narrow your search results based on specific criteria, such as organization, date/time ranges, or geospatial locations.
-
-. *Performance optimization:*
-Reduce the search space before executing queries to improve query execution time and reduce computational overhead.
 
 == Prerequisites
 
--- a/antora.yml
+++ b/antora.yml
@@ -9,6 +9,6 @@ ext:
     HEAD:
       sources:
         docs-server:
-          branches: release/7.6
+          branches: [release/7.6]
       override:
         startPage: server:introduction:intro.adoc


modules/vector-search/pages/pre-filtering-vector-search.adoc
```diff
@@ -8,13 +8,10 @@
...
-Using pre-filtering as part of your vector search offers two key advantages:
+Pre-filtering allows you to execute vector searches over a subset of the vector index. 
+You can search results based on specific criteria, such as organization, date/time ranges, or geospatial locations.
 
-. *Enhanced precision and relevance:*
-Narrow your search results based on specific criteria, such as organization, date/time ranges, or geospatial locations.
-
-. *Performance optimization:*
-Reduce the search space before executing queries to improve query execution time and reduce computational overhead.
+By searching over a subset of the index, you can reduce the computational overhead needed to run your queries.
```
abhinavdangeti: Same comment as `https://github.com/couchbase/docs-server/pull/3769` - this is not a performance optimization and does not reduce the compute.

abhinavdangeti (changes requested):

abhinavdangeti (changes requested):
It's not allowing me to "Request changes" it seems.

modules/vector-search/pages/pre-filtering-vector-search.adoc
RayOffiah: So my question is, why would they do it?

abhinavdangeti: Alright, we're pursuing this conversation on both your commits - so let me try this one more time then -

- Nearest-neighbor (vector) search is a non-exhaustive search technique that obtains the top-K nearest (not necessarily exact) hits for a search vector. The inverted index data structures used for vector search are different from the traditional search data structures - so no overlap.
- When a hybrid search ensues combining elements of traditional search and nearest neighbor search - there need not always be an overlap .. meaning high result-relevancy will become impossible to achieve in certain cases.
- This is where **filtered vector search** becomes a crucial feature, where ..
    - In the "first phase" all the results that qualify the filter (traditional exact search) are determined.
    - In the "second phase", we will run the vector search - except this time we will feed it a selectivity setting that includes all the hits from the first-phase search and top-K is estimated over them.
- "Filtered vector search" helps deliver accuracy when running vector search over a portion of the indexed content and I hope you follow why this is not a performance optimization.

@RayOffiah If I'm for any reason still not making sense, there're several online resources with literally every competitor of ours supporting this feature that you can look up. Here's a nice blog talking about this - https://myscale.com/blog/filtered-vector-search-in-myscale/#filtered-vector-search-is-essential-for-improving-the-accuracy-of-rag-systems 

cc @sarahlwelton

RayOffiah: @abhinavdangeti 

> "Filtered vector search" helps deliver accuracy when running vector search over a portion of the indexed content optimization[…]"

Right, you've now finally answered the question I was originally asking: what is the user benefit from doing this?

Rather than just giving a technical explanation of what's going on, it can sometimes be helpful to supply the users with reasons as to why doing it might be a good idea.

modules/vector-search/pages/pre-filtering-vector-search.adoc
```diff
@@ -8,13 +8,12 @@
 
 == About Pre-filtering
 
-Using pre-filtering as part of your vector search offers two key advantages:
+The Search Service supports pre-filtering on vector search queries. 
+Pre-filtering allows you to execute vector searches over a subset of the vector index, 
+via the means of a filter request that qualifies the subset.
```
sarahlwelton: @RayOffiah , could you make the same changes here that I proposed on https://github.com/couchbase/docs-server/pull/3769?

modules/vector-search/pages/pre-filtering-vector-search.adoc
```diff
@@ -8,13 +8,12 @@
...
 
-. *Enhanced precision and relevance:*
-Narrow your search results based on specific criteria, such as organization, date/time ranges, or geospatial locations.
+Filtered vector searches help deliver accuracy when running vector search over a portion of the indexed content optimization.
```
abhinavdangeti: Again this is _not_ an optimization.

modules/vector-search/pages/pre-filtering-vector-search.adoc
RayOffiah: Can do

abhinavdangeti (approved):

PR #320, opened 2025-02-18
[AV-93910] Document App Services Compatibility setting

dgnicholson:
Eventing will be available to App Services with 7.6.5 (in a follow-up change at the end of Feb). This adds mention of the setting in the Create Function proc, as well as in the modify functions page.

Related PR for App Services is here: https://github.com/couchbase/docs-capella/pull/136


simon-dew (approved):
LGTM 👍

PR #327, opened 2025-03-14
[DOC-9249] Add HASHBYTES Function

rakhi-prathap:
[DOC-9249]

Documentation for hashbytes() function. 

Preview: https://preview.docs-test.couchbase.com/DOC-9249-hashbytes/server/current/n1ql/n1ql-language-reference/metafun.html#hashbytes

[DOC-9249]: https://couchbasecloud.atlassian.net/browse/DOC-9249?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/n1ql/pages/n1ql-language-reference/metafun.adoc
+++ b/modules/n1ql/pages/n1ql-language-reference/metafun.adoc
@@ -714,11 +714,8 @@ By using this function, you can verify or compare data quickly, or protect your 
 
 === Arguments
 
-input:: The data to be hashed, represented as a binary object.
-+
-If the data is not in the binary form, you must convert it to a JSON marshalled value (JSON string) before it can be hashed. 
-You can use the ENCODE_JSON() function to do this conversion. 
-For more information, see xref:n1ql-language-reference/jsonfun.adoc[JSON Functions].     
+input:: A binary object or any {sqlpp} data type.
+The JSON marshalled value of the data is used as the input.
 
 options::  [Optional]
 An object that specifies the hashing algorithm and other options for the function.
--- a/modules/n1ql/pages/n1ql-language-reference/metafun.adoc
+++ b/modules/n1ql/pages/n1ql-language-reference/metafun.adoc
@@ -703,43 +703,53 @@ FROM system:completed_requests;
 ----
 ====
 
+[[hashbytes,HASHBYTES()]]
+
 == HASHBYTES(`input`, [ `options` ])
 
 === Description
 
-This function returns a binary hash value for a given input using a specified hashing algorithm. It accepts both binary and JSON data as inputs and supports various hashing algorithms. 
+This function returns a binary hash value for a given input using a specified hashing algorithm. 
+By using this function, you can protect your data by masking its original form while still allowing verification or comparison. 
 
 === Arguments
 
-input:: A binary object to hash. 
+input:: The data to be hashed, represented as a binary object.
 +
-For non-binary data, you must provide the JSON marshalled value.   
+If the data is not in the binary form, you must convert it to a JSON marshalled value (JSON string) before it can be hashed. 
+You can use the ENCODE_JSON() function to do this conversion. 
+For more information, see xref:n1ql-language-reference/jsonfun.adoc[JSON Functions].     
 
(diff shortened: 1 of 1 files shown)


modules/n1ql/pages/n1ql-language-reference/metafun.adoc
```diff
@@ -703,6 +703,84 @@ FROM system:completed_requests;
...
++
+Some of the algorithms require additional configuration fields. 
+For example, for `crc32`, you need to add a `polynomial` field which can have one of the following values: 
++ 
+--
+* `ieee` (the default value)
+* `castagnoli`
+* `koopman`
+* A valid 32-bit integer, provided either as a JSON number (decimal) or a string that can be parsed as a numeric value (supports hexadecimal with a "0x" prefix)
+--
```
simon-dew: We need to be more explicit in the way the options object is documented. For an example of how we do this, see [_TIMESERIES Function](https://docs.couchbase.com/server/current/n1ql/n1ql-language-reference/timeseries.html).  Here's how I would do it:

```suggestion
options::  [Optional]
An object that specifies the hashing algorithm and other options for the function.
If omitted, the default hashing algorithm is `sha256`.

=== Options

[options="header", cols="1a,3a,1a"]
|===
| Name | Description | Schema

| **algorithm** +
__required__
| Specifies the hashing algorithm to be used.

Supported algorithms are: `crc32`, `md4`, `md5`, `sha224`, `sha256`, `sha384`, `sha512`, `sha512/224`, and `sha512/256`. 
| String

| **polynomial** +
__optional__
| The polynomial to use.
This only applies if the algorithm is `crc32`.
This property may have one of the following values:

* `ieee` (the default value)
* `castagnoli`
* `koopman`
* A valid 32-bit integer, provided either as a JSON number (decimal) or a string that can be parsed as a numeric value (supports hexadecimal with a "0x" prefix)

*Default:* `ieee`
| String or integer
|===
```

modules/n1ql/pages/n1ql-language-reference/metafun.adoc
```diff
@@ -703,6 +703,84 @@ FROM system:completed_requests;
...
+This function returns a binary hash value for a given input using a specified hashing algorithm. 
+By using this function, you can protect your data by masking its original form while still allowing verification or comparison. 
+
+=== Arguments
+
+input:: The data to be hashed, represented as a binary object.
++
+If the data is not in the binary form, you must convert it to a JSON marshalled value (JSON string) before it can be hashed. 
+You can use the ENCODE_JSON() function to do this conversion. 
+For more information, see xref:n1ql-language-reference/jsonfun.adoc[JSON Functions].
```
simon-dew: Looking closely at the original ticket, I'm not sure if this is quite right.

```suggestion
input:: A binary object or any {sqlpp} data type.
The JSON marshalled value of the data is used as the input.
```

That's how I read it anyway. If you're in doubt you can test the input by trying out the function with a number, an object, an array, NULL, and MISSING, and see whether you get valid output each time.

modules/n1ql/pages/n1ql-language-reference/metafun.adoc
```diff
@@ -703,6 +703,84 @@ FROM system:completed_requests;
...
 ====
 
+[[hashbytes,HASHBYTES()]]
+
+== HASHBYTES(`input`, [ `options` ])
+
+=== Description
+
+This function returns a binary hash value for a given input using a specified hashing algorithm. 
+By using this function, you can protect your data by masking its original form while still allowing verification or comparison.
```
simon-dew: Just tweaking the wording to emphasize that it's not just for security.

```suggestion
By using this function, you can verify or compare data quickly, or protect your data by masking its original form while still allowing verification or comparison. 
```

simon-dew (changes requested):
This is great, thanks @rakhi-prathap ! I got picky on the details, especially being precise about the options object

simon-dew (approved):
LGTM 👍

roikatz (approved):
Validated the documentation for this feature and it looks readable and clear and convey the feature main functions.

PR #333, opened 2025-03-25
[DOC-10871] Add OBJECT_CONCAT2 Function

rakhi-prathap:
[DOC-10871]

Added a new OBJECT_CONTACT2 function that takes both plain objects and arrays of objects as inputs. 

Preview: https://preview.docs-test.couchbase.com/DOC-10871-object-concat2/server/current/n1ql/n1ql-language-reference/objectfun.html#fn-obj-concat2

[DOC-10871]: https://couchbasecloud.atlassian.net/browse/DOC-10871?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/n1ql/pages/n1ql-language-reference/objectfun.adoc
+++ b/modules/n1ql/pages/n1ql-language-reference/objectfun.adoc
@@ -109,6 +109,62 @@ SELECT OBJECT_CONCAT({"abc": 1}, {"def": 2}, {"ghi": 3}, {"ghi": 4}, {"ghi": [5,
 ----
 ====
 
+[[fn-obj-concat2,OBJECT_CONCAT2()]]
+== OBJECT_CONCAT2(`expr`, `expr` ...)
+
+=== Description
+
+This function concatenates multiple input objects into a single new object, and requires at least two input objects to work. 
+Unlike <<fn-obj-concat>>, this function supports both plain objects and arrays of objects as arguments. 
+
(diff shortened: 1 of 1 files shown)


dhanyagowrish (approved):

dhanyagowrish (changes requested):
The example provided does not include any arrays of objects as arguments.

Can we please change the example?

An example would be the following query. Which includes an array of objects and an object as the arguments:

`select object_concat2({"a":1}, [{"b":2}, {"c":3}, {"d":4}], {"e":5});`

Thanks

rakhi-prathap:
> The example provided does not include any arrays of objects as arguments.
> 
> Can we please change the example?
> 
> An example would be the following query. Which includes an array of objects and an object as the arguments:
> 
> `select object_concat2({"a":1}, [{"b":2}, {"c":3}, {"d":4}], {"e":5});`
> 
> Thanks

Got it. Have updated the example. 

Also, I've replaced the generic a, b, c objects with a flight-related example to maintain consistency with other examples on the page. Made the change for OBJECT_CONCAT() too.

dhanyagowrish (approved):

RichardSmedley (approved):
LGTM :+1:

PR #334, opened 2025-03-26
[DOC-10857] Add a note to ARRAY_REMOVE()

rakhi-prathap:
DOC-10857

Added a note to the ARRAY_REMOVE function about removing NULL values. 

Preview: https://preview.docs-test.couchbase.com/docs-devex-DOC-10857-update-array-remove/server/current/n1ql/n1ql-language-reference/arrayfun.html#fn-array-remove

[DOC-10857]: https://couchbasecloud.atlassian.net/browse/DOC-10857?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/n1ql/pages/n1ql-language-reference/arrayfun.adoc
+++ b/modules/n1ql/pages/n1ql-language-reference/arrayfun.adoc
@@ -1161,7 +1161,7 @@ A new array with all occurrences of the specified [.var]`val` or multiple [.var]
 If any of the arguments are `MISSING`, then it returns `MISSING`.
 +
 If the first argument is not an array, then it returns `NULL`.
-
++
 [NOTE]
 ====
 This function cannot be used to remove NULL values from an array as it uses an equality predicate when evaluating the array elements to remove. 
--- a/modules/n1ql/pages/n1ql-language-reference/arrayfun.adoc
+++ b/modules/n1ql/pages/n1ql-language-reference/arrayfun.adoc
@@ -1170,7 +1170,9 @@ For example:
 
 [source,sqlpp]
 ----
-SELECT ARRAY a FOR a IN [1, NULL, 2] WHEN a IS NOT NULL END AS filtered_array;
+SELECT ARRAY name FOR name IN ["Ryan", NULL, "Corrine"] 
+WHEN name IS NOT NULL 
+END AS filtered_names;
 ----
 ====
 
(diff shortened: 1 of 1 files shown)


dhanyagowrish (changes requested):
The description of the DOC ticket (DOC-10857) includes the reason why ARRAY_REMOVE() cannot remove NULL values from an array. Can that be added?

Reasoning from the DOC ticket:

> The ARRAY_REMOVE function can't be used to remove NULL values from an array since it uses an equality predicate and NULL doesn't equal NULL.

The internal implementation of the ARRAY_REMOVE() function uses an equality predicate when evaluating the array elements to remove. And since NULL does not equal NULL, the implementation will not remove the NULL values from the array,

rakhi-prathap:
> The description of the DOC ticket ([DOC-10857](https://jira.issues.couchbase.com/browse/DOC-10857)) includes the reason why ARRAY_REMOVE() cannot remove NULL values from an array. Can that be added?
> 
> Reasoning from the DOC ticket:
> 
> > The ARRAY_REMOVE function can't be used to remove NULL values from an array since it uses an equality predicate and NULL doesn't equal NULL.
> 
> The internal implementation of the ARRAY_REMOVE() function uses an equality predicate when evaluating the array elements to remove. And since NULL does not equal NULL, the implementation will not remove the NULL values from the array,

Done. Updated the doc to include this information. 

[DOC-10857]: https://couchbasecloud.atlassian.net/browse/DOC-10857?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

dhanyagowrish (approved):

modules/n1ql/pages/n1ql-language-reference/arrayfun.adoc
```diff
@@ -1157,11 +1157,27 @@ val1, val2, …:: [At least 1 is required] The input value or multiple values to
 
 Output Values::
 A new array with all occurrences of the specified [.var]`val` or multiple [.var]`val` fields removed from the array [.var]`expr`.
-
++
 If any of the arguments are `MISSING`, then it returns `MISSING`.
-
++
 If the first argument is not an array, then it returns `NULL`.
```
RichardSmedley: Is it worth adding a "+" here, maybe,
to indent the note to the same rank as the NULL related content in the Output Values section above?

(I don't know the form for these SQL++ references, there may be good reasons why you wouldn't).

rakhi-prathap: Yes, makes sense. Have indented the note to the same level as the previous line.

RichardSmedley (approved):
:+1:

PR #347, opened 2025-05-16
DOC-13223 sizing guidelines, note on residentRatio

TimLFletcher:
Added a note to signal that low residentRatio values can lead to issues with indexing.

Linked to the server docs change here that gives guardrails info:

https://github.com/couchbase/docs-server/pull/3808

### DIFF
--- a/modules/n1ql/pages/n1ql-language-reference/index-partitioning.adoc
+++ b/modules/n1ql/pages/n1ql-language-reference/index-partitioning.adoc
@@ -532,7 +532,7 @@ Non-array fields will be ignored.
 | `50`
 |===
 
-WARNING: We recommend keeping residentRatio above 10% to avoid issues like index build failures.
+NOTE: We recommend keeping residentRatio above 10 to avoid issues like index build failures.
 
 To provide sizing estimation, you can use a command similar to the following examples.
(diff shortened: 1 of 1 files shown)


osfameron:
Here's what the bot has to say:
image

So that's a 50% success... 🤦

osfameron (approved):
Correctly based off release/7.6 👍

osfameron:
Recommend editing the PR name to something like "DOC-13223 sizing guidelines, note on residentRatio"

rao-shwe (changes requested):
Change to:

NOTE: Couchbase recommends setting the `residentRatio` property value over 10 to avoid issues, for example, index build failures.


- Instead of "we", the generic technical writing guidelines suggest to use the company's brand name.
- residentRatio is an object property, so it is better to use the complete nomenclature, which is "residentRatio property".
- "over" goes well with conversational style of writing.
- "10" is a value, so it would be better to define what "10" represents.
- "like" becomes quite a casual word for a NOTE. Also, it would mean that we are directing the user's attention to "index build failures". But "for example" eases the information about the failure.

rao-shwe (approved):
All good!

PR #367, opened 2025-06-11
Removing single reference to memcached buckets

ggray-cb:
Associated with DOC-12483, which removes all references t the now removed memcached buckets.

### DIFF
--- a/modules/eventing/pages/eventing-faq.adoc
+++ b/modules/eventing/pages/eventing-faq.adoc
@@ -47,7 +47,7 @@ This means that the event entry points of *OnUpdate* and *OnDelete* to the Event
 
 * Can a Function listen to all changes across Couchbase Server?
 +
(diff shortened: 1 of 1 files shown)


modules/eventing/pages/eventing-faq.adoc
```diff
@@ -48,7 +48,6 @@ This means that the event entry points of *OnUpdate* and *OnDelete* to the Event
 * Can a Function listen to all changes across Couchbase Server?
 +
 A defined Function listens only to changes that are exposed using the DCP for the source collection (Couchbase and Ephemeral) in the data-nodes.
```
RichardSmedley: The "(Couchbase and Ephemeral)" is probably also not necessary, as it is no longer a positive mention in contrast to the never-available (in eventing) memcached buckets.

hyunjuV (approved):
lgtm

PR #374, opened 2025-07-04
[DOC-12695] New SQL++ Statement - USING AI

rakhi-prathap:
This PR adds documentation for the new `USING AI` SQL++ statement. 

Jira: DOC-12695
Preview: [USING AI](https://preview.docs-test.couchbase.com/docs-devex-DOC-12695-natural-lang-requests/server/current/n1ql/n1ql-language-reference/using-ai.html)
Preview credentials:  [Preview docs for internal review](https://couchbasecloud.atlassian.net/wiki/x/dYF_dQ#Preview-docs-for-internal-review)


modules/n1ql/pages/n1ql-language-reference/using-ai.adoc
```diff
@@ -0,0 +1,259 @@
...
+----
+
+image::n1ql-language-reference/using-ai.png["Syntax diagram: refer to source code listing", align=left]
+
+natural-language-request:: A natural language prompt that you want to convert into a SQL++ query.
+
+options::
+
+[Optional] An object specifying additional <<optional-parameters,options>> to guide the statement execution.
+If specified, these options override the default `natural_orgId`, `natural_cred` and `natural_context` request parameters.
```
GauravJayaraj: typo: 
~natural_orgId~ -> natural_orgid

modules/n1ql/pages/n1ql-language-reference/using-ai.adoc
```diff
@@ -0,0 +1,259 @@
...
+If security is a concern, consider using an HTTPS connection. 
+
+== Examples
+
+.Natural language prompt with the default parameters
+====
+.Request
+[source, console]
+----
+cbq> \set -natural_context travel-sample.inventory.hotel;
```
GauravJayaraj: 1) When I copy and paste this in cbq shell, it fails. Can you make the edits required so a first-time user can simply walk through the examples?

2) for 
`\set -natural_cred "username@couchbase.com";`
and 
`\set -natural_orgId 7a99d00c-####-4b39-bc72-############;` <br>Can you use a notation that indicates this would have to be substituted, maybe:
`\set -natural_orgId <7a99d00c-####-4b39-bc72-############>;`
`\set -natural_cred "<username@couchbase.com>";`

3) Would we also need to explain what orgid and creds are? Where can a user look them up?

GauravJayaraj (changes requested):

GauravJayaraj (changes requested):
There's another section you can add as a note, which would come in handy for when the following concern is raised by a user.

An advanced SQL user might recognise that as part of changes for USING AI statements support, **"ai"** would now be a keyword. Which would mean that applications with queries that used ai as a field name would now fail, for example, SELECT ai FROM xyz; 

But this is not the case; special handling is done so "ai" is not a keyword and is only recognized as one for the USING AI statement. 

To summarise, previous queries that used "ai" as a field name would not fail after the new syntax introduction.

modules/n1ql/pages/n1ql-language-reference/using-ai.adoc
rakhi-prathap: Both seem to work, but following our Slack discussion, I've updated all occurrences to `natural_orgid`.

modules/n1ql/pages/n1ql-language-reference/using-ai.adoc
rakhi-prathap: Have addressed them all. 
- Updated the examples.
- Explicitly mentioned that users need to replace `natural_cred` and `natural_orgid` with their Capella credentials and org ID, respectively. 
- Improved the descriptions for `orgid` and `creds`.

rakhi-prathap:
> There's another section you can add as a note, which would come in handy for when the following concern is raised by a user.
> 
> An advanced SQL user might recognise that as part of changes for USING AI statements support, **"ai"** would now be a keyword. Which would mean that applications with queries that used ai as a field name would now fail, for example, SELECT ai FROM xyz;
> 
> But this is not the case; special handling is done so "ai" is not a keyword and is only recognized as one for the USING AI statement.
> 
> To summarise, previous queries that used "ai" as a field name would not fail after the new syntax introduction.

Done. Added this as a note.

GauravJayaraj (approved):
Looks good to me. Approving for merge 👍

modules/n1ql/pages/n1ql-language-reference/using-ai.adoc
```diff
@@ -0,0 +1,297 @@
...
+
+natural-language-request:: A natural language prompt that you want to convert into a SQL++ query.
+
+options::
+
+[Optional] A JSON object specifying additional <<optional-parameters,options>> to guide the statement execution.
+By default, the statement uses the `natural_orgid`, `natural_cred`, and `natural_context` request parameters to determine the organization ID, credentials, and keyspaces for the request.
+However, you can override these defaults by using the <<optional-parameters,options>> object.
+ 
+//TODO: Add links to `natural_orgid`, `natural_cred` and `natural_context` when available in the 8.0 documentation.
```
RayOffiah, line 50: This is coming in a later release, yes?

RayOffiah (approved):

modules/n1ql/pages/n1ql-language-reference/using-ai.adoc
rakhi-prathap: These are all part of the Morpheus 8.0 release - same as this doc. However, the documentation for that part is not ready yet. I'll add the links when they are available (will create another PR to add them).

PR #275, opened 2024-10-07
DOC-12434: DDL for GSI Indexes with a Vector field

simon-dew:
Docs issues: DOC-12434 and DOC-12436

This PR updates the DDL for both hyperscale vector indexes and composite vector indexes.

Docs preview:
- Overview topics
  * [Primary and Secondary Index Reference](https://preview.docs-test.couchbase.com/docs-devex-DOC-12434/server/current/indexes/indexing-overview.html)
  * [Types of Primary and Secondary Index](https://preview.docs-test.couchbase.com/docs-devex-DOC-12434/server/current/indexes/indexing-and-query-perf.html)
- Hyperscale vector indexes
  * [CREATE VECTOR INDEX](https://preview.docs-test.couchbase.com/docs-devex-DOC-12434/server/current/n1ql/n1ql-language-reference/createvectorindex.html)
   * [BUILD INDEX](https://preview.docs-test.couchbase.com/docs-devex-DOC-12434/server/current/n1ql/n1ql-language-reference/build-index.html)
  * [ALTER VECTOR INDEX](https://preview.docs-test.couchbase.com/docs-devex-DOC-12434/server/current/n1ql/n1ql-language-reference/altervectorindex.html)
  * [DROP VECTOR INDEX](https://preview.docs-test.couchbase.com/docs-devex-DOC-12434/server/current/n1ql/n1ql-language-reference/dropvectorindex.html)
- Composite vector indexes
   * [CREATE INDEX](https://preview.docs-test.couchbase.com/docs-devex-DOC-12434/server/current/n1ql/n1ql-language-reference/createindex.html)
   * [BUILD INDEX](https://preview.docs-test.couchbase.com/docs-devex-DOC-12434/server/current/n1ql/n1ql-language-reference/build-index.html)
   * [DROP INDEX](https://preview.docs-test.couchbase.com/docs-devex-DOC-12434/server/current/n1ql/n1ql-language-reference/dropindex.html)

Credentials: [Preview docs for internal review](https://couchbasecloud.atlassian.net/wiki/x/dYF_dQ#Preview-docs-for-internal-review)

> [!NOTE]
> Many of the links in the preview docs will not work until the following PR is merged:
> * #325

PR #406, opened 2025-08-26
DOC-12940: GSI Vector Index not supported on Windows

simon-dew:
Docs issue: DOC-12940


simon-dew:
Merging as-is so that it can be reviewed as part of https://github.com/couchbaselabs/docs-devex/pull/275

PR #412, opened 2025-09-03
[DOC-13464] Updates to the USING AI statement

rakhi-prathap:
Doc tickets: DOC-12770, DOC-13464, DOC-13492

This PR includes the following updates: 

- [USING AI statement](https://preview.docs-test.couchbase.com/docs-devex-DOC-13464-using-ai-fts/server/current/n1ql/n1ql-language-reference/using-ai.html): 
   - A new `output` field that supports the following values - `sql`, `jsudf`, `ftssql`
   - A note to mention EXPLAIN and ADVISE can be used as prefixes. 
   - Some new examples. 
   - Some language and formatting edits.
- [ADVISE statement](https://preview.docs-test.couchbase.com/docs-devex-DOC-13464-using-ai-fts/server/current/n1ql/n1ql-language-reference/advise.html): 
   - Updated EBNF grammar and doc to include USING AI.

[Doc preview credentials](https://couchbasecloud.atlassian.net/wiki/x/dYF_dQ#Preview-docs-for-internal-review)


modules/n1ql/pages/n1ql-language-reference/using-ai.adoc
```diff
@@ -294,4 +337,258 @@ curl -s -d "natural_cred=username@example.com:${p}" \
...
+        "resultCount": 1,
+        "resultSize": 25,
+        "serviceLoad": 3,
+        "naturalLanguageProcessingTime": "2.889671542s"
+    }
+}
+----
+====
+
+[[example-7]]
```
GauravJayaraj: There is one thing we missed for execute option.
https://github.com/couchbaselabs/docs-devex/pull/412/files#diff-cbf3cd2f289fbfffb2d01ff505bc5ed748dd7392b753b8b1fe97691a515be7a9R108 

The default value for execute option is TRUE, but the behaviour is only limited to SELECT (and EXPLAIN/ADVISE if prefixed). Which means when the generated statement is  an INSERT/UPDATE/DELETE/UPSERT or a CREATE FUNCTION statement even though execute is true we only return the generate statement.
This was done to give more control to the user and not allow using ai statements to create functions or insert/update/delete/upsert documents which users cannot undo or may find difficult track/manage later. For generated statements other than select, if a user wishes to execute them they will have to do it themselves; a new request with the generated statement as the statement.

We can mention the same over here or at https://github.com/couchbaselabs/docs-devex/pull/412/files#diff-cbf3cd2f289fbfffb2d01ff505bc5ed748dd7392b753b8b1fe97691a515be7a9R127 for the jsudf option too. 

We can maybe add an example where the generated statement is an INSERT and highlight that the execute option will not be in effect.

GauravJayaraj (changes requested):

modules/n1ql/pages/n1ql-language-reference/using-ai.adoc
rakhi-prathap: Done. Have made the necessary changes.

GauravJayaraj (approved):
Looks good to me:)

iamfortune (approved):
LGTM

PR #419, opened 2025-09-12
NO ISSUE: Consolidate the vector index files

simon-dew:


simon-dew:
Minor fixes. To be reviewed as part of DOC-12294

PR #422, opened 2025-09-15
DOC-13545 GSI Vector Index Similarity internally uses squared variant for L2 and Euclidean

ggray-cb:
* Added note to Euclidean description that it uses Euclidean Squared Distance instead behind the scenes.
* Changed "FTS Vector Index" to "Search Vector Index" which is PM's current nomenclature.

Preview URL:
https://preview.docs-test.couchbase.com/docs-server-DOC-12565_vector_search_concepts/server/current/vector-index/vectors-and-indexes-overview.html#euclidean

### DIFF
--- a/modules/vector-index/pages/vectors-and-indexes-overview.adoc
+++ b/modules/vector-index/pages/vectors-and-indexes-overview.adoc
@@ -127,7 +127,13 @@ Use this method when the actual distance of the vectors and their magnitudes are
 This method is useful if the distance between vectors represents a real-world value.
  
 image::euclidean-distance-example.svg["Three-dimensional plot showing two vectors with points along each vector joined by dotted lines, indicating the summing of corresponding points."]
- 
+
+NOTE: When you select Euclidean Distance or L2 as the metric for a vector index, Couchbase {product-name} internally uses the <<#euclidean-squared>> metric (explained in the sext section) to perform vector comparisons. 
+This approach improves performance because it avoids performing a computationally expensive square root operation.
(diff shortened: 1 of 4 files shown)


simon-dew (approved):
LGTM 👍

modules/vector-index/pages/vectors-and-indexes-overview.adoc
```diff
@@ -127,21 +127,28 @@ Use this method when the actual distance of the vectors and their magnitudes are
 This method is useful if the distance between vectors represents a real-world value.
  
 image::euclidean-distance-example.svg["Three-dimensional plot showing two vectors with points along each vector joined by dotted lines, indicating the summing of corresponding points."]
- 
+
+NOTE: When you select Euclidean Distance or L2 as the metric for a vector index, Couchbase {product-name} internally uses the <<#euclidean-squared>> metric (explained in the sext section) to perform vector comparisons.
```
simon-dew: I don't think you need to include the hash in an [internal cross reference](https://docs.asciidoctor.org/asciidoc/latest/macros/xref/#internal-cross-references)? No worries if it works

```suggestion
NOTE: When you select Euclidean Distance or L2 as the metric for a vector index, Couchbase {product-name} internally uses the <<euclidean-squared>> metric (explained in the sext section) to perform vector comparisons. 
```

ggray-cb: It does work. Looking it up, it seems like it's not needed. There are a ton of them in the Server docs, and I just copied practie from there. I'll accept the change and try to change how I use them from now on. No telling if they will break someday.

PR #436, opened 2025-10-01
[AV-112236] Replace instances of tfidf with tf-idf

sarahlwelton:
Fixing instances of tfidf to be tf-idf, as per PM. 

Relates to https://github.com/couchbaselabs/docs-devex/pull/435


Rebecca-Martinez007 (approved):
👍

PR #452, opened 2025-10-17
DOC-13644: [Capella] Vector-index sample bucket

simon-dew:
Docs issue: DOC-13644

For Capella only:
- Updates the name of the vector-sample bucket to `color-vector-sample`
- Adds instructions to import `rgb-questions`, which is not included by default
- Updates queries with `USE KEYS` to avoid creating primary indexes unless absolutely necessary &mdash; Query without Indexes is not available by default for paid service plans
- Updates some existing queries using the `rgb` collection, because the ID for each document has no `#` at the start, unlike when you import from`color_data_2vectors.zip`  

Document previews:
- [Vector Search Using Hyperscale Vector Indexes › Examples on this Page](https://preview.docs-test.couchbase.com/docs-devex-DOC-13644-color-vector-index/cloud/vector-index/hyperscale-vector-index.html#examples-on-this-page)
- [Filtered Search Using Composite Vector Indexes › Examples on this Page](https://preview.docs-test.couchbase.com/docs-devex-DOC-13644-color-vector-index/cloud/vector-index/composite-vector-index.html#examples-on-this-page)

... and many others, these are just the pages with the most changes.

### DIFF
--- a/modules/vector-index/pages/composite-vector-index.adoc
+++ b/modules/vector-index/pages/composite-vector-index.adoc
@@ -52,6 +52,19 @@ See xref:vector-index:vectors-and-indexes-overview.adoc#vector_similarity[Vector
 
 include::vector-search:partial$download-sample-partial.adoc[]
 
+include::vector-search:partial$download-sample-partial.adoc[]
+
+In addition, to get the best results with the examples on this page, download the following sample dataset:
+
+https://cbc-remote-execution-examples-prod.s3.amazonaws.com/color_data_2vectors.zip[Download color_data_2vectors.zip]
+
+When you have downloaded `color_data_2vectors.zip`, unzip it, then xref:guides:load.adoc[import] the `rgb_questions.json` file with the following settings:
+
+* Use the `color-vector-sample` bucket.
+* Use the `color` scope.
+* Create a new target collection called `rgb-questions` -- with a hyphen `-`, not an underscore `&lowbar;`.
+* To set your document keys, use the value of the `id` field from the JSON document.
+
 [#create-index]
 
 include::vector-index:partial$create-composite-index-capella-ui.adoc[opts=optional]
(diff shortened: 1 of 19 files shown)


modules/vector-index/pages/composite-vector-index.adoc
```diff
@@ -52,6 +52,19 @@ See xref:vector-index:vectors-and-indexes-overview.adoc#vector_similarity[Vector
...
+In addition, to get the best results with the examples on this page, download the following sample dataset:
+
+https://cbc-remote-execution-examples-prod.s3.amazonaws.com/color_data_2vectors.zip[Download color_data_2vectors.zip]
+
+When you have downloaded `color_data_2vectors.zip`, unzip it, then xref:guides:load.adoc[import] the `rgb_questions.json` file with the following settings:
+
+* Use the `color-vector-sample` bucket.
+* Use the `color` scope.
+* Create a new target collection called `rgb-questions` -- with a hyphen `-`, not an underscore `&lowbar;`.
+* To set your document keys, use the value of the `id` field from the JSON document.
```
sarahlwelton, line 66: @simon-dew This is not necessary for Capella. You can install the sample data just like travel-sample from the *Import* tab. Please correct this - they added the sample to Capella for exactly this reason:

image

modules/vector-index/pages/hyperscale-vector-index.adoc
```diff
@@ -63,6 +63,17 @@ See xref:vector-index:vectors-and-indexes-overview.adoc#vector_similarity[Vector
...
+In addition, to get the best results with the examples on this page, download the following sample dataset:
+
+https://cbc-remote-execution-examples-prod.s3.amazonaws.com/color_data_2vectors.zip[Download color_data_2vectors.zip]
+
+When you have downloaded `color_data_2vectors.zip`, unzip it, then xref:guides:load.adoc[import] the `rgb_questions.json` file with the following settings:
+
+* Use the `color-vector-sample` bucket.
+* Use the `color` scope.
+* Create a new target collection called `rgb-questions` -- with a hyphen `-`, not an underscore `&lowbar;`.
+* To set your document keys, use the value of the `id` field from the JSON document.
```
sarahlwelton, line 75: Same thing here - the sample can be installed from the *Import* tab, none of this is necessary.

sarahlwelton (changes requested):
I can still pop back on to approve this if you want it to go in today, but we need to fix this part.

modules/vector-index/pages/composite-vector-index.adoc
simon-dew: No you've missed the point. The color-vector-sample that is provided via the UI **does not include** the rgb-questions collection. You need to install that by hand. Hence these instructions.

simon-dew: Sorry if I didn't make this clear in the PR description

sarahlwelton (approved):

PR #471, opened 2025-11-14
[DOC-13700] Update EXTRACTDDL() to add new flags 8.0.1

rakhi-prathap:
Adds two new flags to the EXTRACTDDL function. 

Jira: DOC-13700
Docs preview: [EXTRACTDDL()](https://preview.docs-test.couchbase.com/docs-devex-DOC-13700-update-extract-ddl/server/current/n1ql/n1ql-language-reference/metafun.html#extract-ddl)
Preview credentials:  [Credentials for internal review](https://couchbasecloud.atlassian.net/wiki/x/dYF_dQ#Preview-docs-for-internal-review)

### DIFF
--- a/modules/n1ql/pages/n1ql-language-reference/metafun.adoc
+++ b/modules/n1ql/pages/n1ql-language-reference/metafun.adoc
@@ -439,7 +439,7 @@ The function supports the following statements:
 * CREATE COLLECTION
 * CREATE INDEX
 * CREATE SEQUENCE
-* CREATE FUNCTION [.status]#Couchbase Server 8.0.1#
+* CREATE OR REPLACE FUNCTION [.status]#Couchbase Server 8.0.1#
 * PREPARE [.status]#Couchbase Server 8.0.1#
 
 NOTE: To execute this function, you must have the `query_system_catalog` role. 
@@ -468,10 +468,10 @@ __optional__
 
 Accepts either a numeric value or an array of strings, but not both. 
 
-[options="header", cols="1a,1a,1a"]
+[options="header", cols="3a,2a,1a"]
 !===
 
-! Statement ! String Value ! Numeric Value
+! Statement ! String ! Number
 
 ! CREATE BUCKET
 ! `"bucket"` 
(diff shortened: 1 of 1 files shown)


Sohoxic (commented):
This looks good for the function and prepareds flag but just a suggestion if you could add another example where we donot specify the flags, meaning an empty "" which will give back all the DDL statements(create bucket/scope/collection/index/sequence/function/prepareds). Also if you could add examples for each of the flag separately like we have for index:

`SELECT extractddl("travel-sample",{"flags":["index"]});`


Else everything looks fine to me!

modules/n1ql/pages/n1ql-language-reference/metafun.adoc
```diff
@@ -438,6 +439,8 @@ The function supports the following statements:
 * CREATE COLLECTION
 * CREATE INDEX
 * CREATE SEQUENCE
+* CREATE FUNCTION [.status]#Couchbase Server 8.0.1#
```
simon-dew: For consistency


```suggestion
* CREATE OR REPLACE FUNCTION [.status]#Couchbase Server 8.0.1#
```

modules/n1ql/pages/n1ql-language-reference/metafun.adoc
simon-dew: Maybe make third column narrower?


```suggestion
[options="header", cols="3a,2a,1a"]
```

simon-dew:
```suggestion
! Statement ! String ! Number
```

simon-dew (changes requested):

rakhi-prathap:
> This looks good for the function and prepareds flag but just a suggestion if you could add another example where we donot specify the flags, meaning an empty "" which will give back all the DDL statements(create bucket/scope/collection/index/sequence/function/prepareds). Also if you could add examples for each of the flag separately like we have for index:
> 
> `SELECT extractddl("travel-sample",{"flags":["index"]});`
> 
> Else everything looks fine to me!

Added examples to show how to extract DDLs from all buckets and how to extract all supported DDL statements from a specific bucket by omitting flags. 

Also, since we already have 6 examples for this function, will leave it at that and skip adding examples for each individual flag.

Sohoxic (approved):
Looks good to me!

PR #473, opened 2025-11-14
[AV-115725] Add new AI Functions Error to the SQL Error Reference

Rebecca-Martinez007:
Add new AI Functions Error to the SQL Error Reference

Local Images: 
Screenshot 2025-11-17 at 11 58 31 AM
Screenshot 2025-11-17 at 11 58 47 AM

PR #477, opened 2025-11-21
[AV-115725] Add new AI Functions Error to the SQL Error Reference (#473)

dgnicholson:
* Add new error to the SQL error reference

* Small nit

* Include the example for future changes


Rebecca-Martinez007 (approved):
Ready for GA.

PR #516, opened 2025-12-09
[DOC-13795] Feedback on Install Couchbase Server

rakhi-prathap:
PR to update the dev tutorial `docker run` command to use the correct version tag - 8.0.0.

Jira: DOC-13795
Preview: [Developer Tutorial - Install Couchbase Server](https://preview.docs-test.couchbase.com/docs-devex-DOC-13795-install-server-feedback/server/current/tutorials/install-couchbase-server.html#docker)

### DIFF
--- a/modules/tutorials/pages/install-couchbase-server.adoc
+++ b/modules/tutorials/pages/install-couchbase-server.adoc
@@ -34,7 +34,7 @@ To download Couchbase Server as a Docker instance:
 +
 [source, sh]
 ----
-docker run -d --name db -p 8091-8096:8091-8096 -p 11210-11211:11210-11211 \
+docker run -t --name db -p 8091-8096:8091-8096 -p 11210-11211:11210-11211 \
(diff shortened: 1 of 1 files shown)


RichardSmedley (approved):
LGTM
But a future improvement would be to use an attribute, so that the latest patch release is in all of these snippets automatically.
See the SDKs for examples:
https://github.com/couchbase/docs-sdk-java/blob/release/3.10/modules/hello-world/pages/start-using-sdk.adoc?plain=1#L159

rakhi-prathap:
> LGTM But a future improvement would be to use an attribute, so that the latest patch release is in all of these snippets automatically. See the SDKs for examples: https://github.com/couchbase/docs-sdk-java/blob/release/3.10/modules/hello-world/pages/start-using-sdk.adoc?plain=1#L159

Ah yes, makes sense to have an attribute. Will create a separate change request for this improvement. Thanks Richard.

PR #552, opened 2026-01-12
[DOC-10255] Date Functions - Inaccurate Example

rakhi-prathap:
Jira: DOC-10255

This ticket was aimed at fixing an example for `DATE_DIFF_STR()` that showed `BETWEEN` being used with wilcards, and wildcards only work with `LIKE`. 

However, on closer review, the whole example didn't really seem to apply to this function, so I removed it along with the trailing note. 

Preview: [DATE_DIFF_STR > Examples](https://preview.docs-test.couchbase.com/docs-devex-DOC-10255-date-functions-bug/server/current/n1ql/n1ql-language-reference/datefun.html#examples-9)


simon-dew (approved):
LGTM 👍

PR #563, opened 2026-01-21
[DOC-13222]: Feedback on Reserved Words | Couchbase Docs

RayOffiah:
Changed the layout of the reserved words list. 

They're now listed in columns flowing top to bottom.

Hakim should review as he might know of a better way to do this, without embedding CSS code on the AsciiDoc page.

Preview URL:
https://preview.docs-test.couchbase.com/docs-devex-DOC-13222/release/8.0/Feedback-on-Reserved-Words--Couchbase-Docs/server/current/n1ql/n1ql-language-reference/reservedwords.html

### DIFF
--- a/modules/n1ql/pages/n1ql-language-reference/reservedwords.adoc
+++ b/modules/n1ql/pages/n1ql-language-reference/reservedwords.adoc
@@ -33,282 +33,268 @@ CREATE INDEX myindex ON default(`index`) USING GSI;
 
 The following keywords are reserved and cannot be used as unescaped identifiers:
 
-[cols=6*]
-|===
-| _INDEX_CONDITION
-| _INDEX_KEY
-| xref:n1ql-language-reference/advise.adoc[ADVISE]
-| xref:n1ql-language-reference/selectclause.adoc#all[ALL]
-| ALTER
-| ANALYZE
 
-| xref:n1ql-language-reference/logicalops.adoc#logical-op-and[AND]
-| xref:n1ql-language-reference/collectionops.adoc#collection-op-any[ANY]
-| xref:n1ql-language-reference/collectionops.adoc#array[ARRAY]
-| xref:n1ql-language-reference/from.adoc#section_ax5_2nx_1db[AS]
-| ASC
-| AT
+++++
+<style>
+.card-row.six-column-row .column {
+  flex-basis: 10%;
+  padding-right: 1rem;
+  padding-left: 1rem;
+  margin-top: 10px
 
-| BEGIN
-| xref:n1ql-language-reference/comparisonops.adoc#between[BETWEEN]
-| xref:n1ql-language-reference/datatypes.adoc#datatype-binary[BINARY]
-| xref:n1ql-language-reference/datatypes.adoc#datatype-boolean[BOOLEAN]
-| BREAK
-| BUCKET
+}
+</style>
+<div class="card-row six-column-row">
 
-| BUILD
-| BY
-| CACHE
-| CALL
-| xref:n1ql-language-reference/conditionalops.adoc[CASE]
-| CAST
+++++
 
-| CLUSTER
-| COLLATE
-| COLLECTION
-| COMMIT
-| xref:n1ql:n1ql-language-reference/set-transaction.adoc[COMMITTED]
-| CONNECT
+[.column]
+_INDEX_CONDITION +
+_INDEX_KEY +
+xref:n1ql-language-reference/advise.adoc[ADVISE] +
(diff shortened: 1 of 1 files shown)


simon-dew (changes requested):
I don't think this looks any better to be honest:

1. For some reason the columns are broken into two unequal sections.
2. The widths of the columns are different in the two sections.
3. The second section only has two columns, wasting a lot of space on the right hand side.
4. The ends of the columns aren't balanced.

I also question the OP's opinion that the "natural way to list words in a table or index is vertically" — citation needed 🤷 

Going back to the original ticket, the OP mentions twice about "initial letter grouping ... group them together for easier reading/locating and appearing in the right-side Table of Contents" and I think this is the ask I would take from this ticket. Break up the list of reserved words with level 3 headings for **Symbol**, **A**, **B**, **C**, and so on.

modules/n1ql/pages/n1ql-language-reference/reservedwords.adoc
```diff
@@ -33,282 +33,274 @@ CREATE INDEX myindex ON default(`index`) USING GSI;
...
 
-[cols=6*]
-|===
-| _INDEX_CONDITION
-| _INDEX_KEY
-| xref:n1ql-language-reference/advise.adoc[ADVISE]
-| xref:n1ql-language-reference/selectclause.adoc#all[ALL]
-| ALTER
-| ANALYZE
+++++
```
osfameron, line 36: Though I'd rather not inline HTML/CSS, it's convenient while trialing a new layout 🤷 

The time to inline a UI solution is arguably when it becomes popular enough that we want to *copy* it to another place.
Is this the first occurrence of this pattern, or is it copied from elsewhere?

modules/n1ql/pages/n1ql-language-reference/reservedwords.adoc
```diff
@@ -33,282 +33,274 @@ CREATE INDEX myindex ON default(`index`) USING GSI;
...
+<style>
+.card-row.six-column-row .column {
+  flex-basis: 10%;
```
osfameron, line 39: Might be worth looking at https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/column-count ?

If that works, then I think it avoids having to mark up the columns explicitly.  
Perhaps a column-count of auto, with a reasonable column-width would make the most sense for responsive layout?

simon-dew: No disrespect intended to the original author of DOC-13222, but I think that ticket is a case where the OP has a genuine issue with the page, but rather than accepting their first proposed solution, we need to work out what the underlying issue is and fix that.

The OP talks twice about grouping the keywords together alphabetically, and even gives an example of what they want with the keywords grouped under headings for each letter. I think that's the solution we should be focussing on, whether the keywords are vertical or horizontal or what.

modules/n1ql/pages/n1ql-language-reference/reservedwords.adoc
RayOffiah: The pattern is copied from the card layout, but I'm not sure it would see much use outside of this page.

modules/n1ql/pages/n1ql-language-reference/reservedwords.adoc
RayOffiah: > Might be worth looking at https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/column-count ?
> 
> If that works, then I think it avoids having to mark up the columns explicitly.
> Perhaps a column-count of auto, with a reasonable column-width would make the most sense for responsive layout?

I had no idea you could do that!

RayOffiah: > The OP talks twice about grouping the keywords together alphabetically, and even gives an example of what they want with the keywords grouped under headings for each letter. I think that's the solution we should be focussing on, whether the keywords are vertical or horizontal or what.

Mmmm. 🤔

He said:

> If you insist on writing the words horizontally, then at least group them together for easier reading/locating and appearing in the right-side Table of Contents, such as:

which I sort of took that as his second choice:

His first choice was:

> The natural way to list words in a table or index is vertically, not horizontally (as shown). 

Something like this:

https://docs.oracle.com/javase/tutorial/java/nutsandbolts/_keywords.html

but avoiding this:

> They look horrible with a blank space (vertically) between words, only add a space between the initial letter grouping.

But his second choice _would_ give him navigation. I'll make the change then you can have a look. 👍🏾

PR #564, opened 2026-01-22
[DOC-13927] Improve Array and Bitwise Functions

rakhi-prathap:
Made the following changes to declutter the TOC and enhance readability: 

- Updated function titles to display only the function name.
- Added a new **Syntax** section to each function, containing the full function definition (name + arguments).
- Moved the existing **Arguments** section to be a sub-section within **Syntax**.

Previews:
- [Array Functions](https://preview.docs-test.couchbase.com/docs-devex-DOC-13927-aggr-array-bitwise-functions/server/current/n1ql/n1ql-language-reference/arrayfun.html)
- [Bitwise Functions](https://preview.docs-test.couchbase.com/docs-devex-DOC-13927-aggr-array-bitwise-functions/server/current/n1ql/n1ql-language-reference/bitwisefun.html)

### DIFF
--- a/modules/n1ql/pages/n1ql-language-reference/arrayfun.adoc
+++ b/modules/n1ql/pages/n1ql-language-reference/arrayfun.adoc
@@ -6,13 +6,12 @@
 {description}
 
 [[fn-array-agg,ARRAY_AGG()]]
-== ARRAY_AGG
+== ARRAY_AGG()
 
-=== Description
-This function returns an array of the non-`MISSING` group values in the input [.var]`expr`, including `NULL` values.
+Returns an array of the non-`MISSING` group values in the input `expr`, including `NULL` values.
 
 === Syntax
-[subs=normal]
+
 ....
 ARRAY_AGG(expr)
 ....
@@ -23,7 +22,7 @@ expr:: [Required] The group of elements you wish to output in an array.
 === Return Values
 An array of non-MISSING values.
 
-If the input [.var]`expr` is `MISSING` or if one of the elements in the array is `MISSING`, then it returns `MISSING`.
+If the input `expr` is `MISSING` or if one of the elements in the array is `MISSING`, then it returns `MISSING`.
 
 === Example
 ====
@@ -53,15 +52,14 @@ SELECT ARRAY_AGG( [ "abc", 1, NULL ] ) AS array_aggregate;
 ====
 
 [[fn-array-append,ARRAY_APPEND()]]
-== ARRAY_APPEND
+== ARRAY_APPEND()
 
-=== Description
-This function takes an array [.var]`expr` and one or more [.var]`val` arguments to return a new array with the specified [.var]`val` argument(s) appended.
+This function takes an array `expr` and one or more `val` arguments to return a new array with the specified `val` argument(s) appended.
 
 It requires a minimum of two arguments and returns an error if there are fewer.
 
 === Syntax
-[subs=normal]
+
 ....
 ARRAY_APPEND(expr, val1, val2, ...)
 ....
@@ -72,15 +70,15 @@ expr:: [Required] The array to be appended to.
 val1, val2, ...:: [At least 1 is required] The values to be appended.
 
 === Return Values
-A new array with the specified [.var]`val` argument(s) appended.
+A new array with the specified `val` argument(s) appended.
 
 If either of the input argument types are `MISSING`, then it returns `MISSING`.
 
 If either of the input argument types are `NULL`, then it returns `NULL`.
 
-If the [.var]`expr` argument is not an array, then it returns `NULL`.
+If the `expr` argument is not an array, then it returns `NULL`.
 
-If the [.var]`expr` is in the `WHERE` clause of a partial index, this function lists the expressions that are implicitly covered.
+If the `expr` is in the `WHERE` clause of a partial index, this function lists the expressions that are implicitly covered.
 
 === Example
 ====
@@ -109,13 +107,12 @@ SELECT ARRAY_APPEND( [ 1 , 2 ] , 3 , 4) AS numbers;
 ====
 
 [[fn-array-avg,ARRAY_AVG()]]
-== ARRAY_AVG
+== ARRAY_AVG()
 
-=== Description
-This function takes an array [.var]`expr` as an argument and returns the arithmetic mean (average) of all the non-`NULL` number values in the array, or `NULL` if there are no such values.
+This function takes an array `expr` as an argument and returns the arithmetic mean (average) of all the non-`NULL` number values in the array, or `NULL` if there are no such values.
 
 === Syntax
-[subs=normal]
+
 ....
 ARRAY_AVG(expr)
 ....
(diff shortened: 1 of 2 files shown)


modules/n1ql/pages/n1ql-language-reference/arrayfun.adoc
```diff
@@ -6,18 +6,24 @@
 {description}
 
 [[fn-array-agg,ARRAY_AGG()]]
-== ARRAY_AGG([.var]`expr`)
+== ARRAY_AGG
 
 === Description
```
simon-dew: Let's consider getting rid of the **Description** heading for all functions. (Keep the description text of course.) It doesn't really add any value, and it triggers the Vale stacked headings rule.

modules/n1ql/pages/n1ql-language-reference/arrayfun.adoc
```diff
@@ -6,18 +6,24 @@
...
+....
+
+==== Arguments
 expr:: [Required] The group of elements you wish to output in an array.
 
 === Return Values
 An array of non-MISSING values.
 
-If the input [.var]`expression` is `MISSING` or if one of the elements in the array is `MISSING`, then it returns `MISSING`.
+If the input [.var]`expr` is `MISSING` or if one of the elements in the array is `MISSING`, then it returns `MISSING`.
```
simon-dew: Get rid of `[.var]` throughout. It's left over from the original import from DITA to AsciiDoc, and we don't use it for anything.

modules/n1ql/pages/n1ql-language-reference/arrayfun.adoc
```diff
@@ -47,17 +53,23 @@ SELECT ARRAY_AGG( [ "abc", 1, NULL ] ) AS array_aggregate;
 ====
 
 [[fn-array-append,ARRAY_APPEND()]]
-== ARRAY_APPEND([.var]`expr`, [.var]`val1`, [.var]`val2`, …)
+== ARRAY_APPEND
```
simon-dew: In the rest of the SQL++ documentation we have developed a sort of convention for referring to functions with brackets, even if we're not specifying any arguments. For example, see [Aggregate Functions](https://docs.couchbase.com/server/current/guides/group-agg.html#aggregate-functions):

> To take multiple values from documents, perform calculations, and return a single value as the result, use an aggregate function, such as [AVG()](https://docs.couchbase.com/server/current/n1ql/n1ql-language-reference/aggregatefun.html#avg), [COUNT()](https://docs.couchbase.com/server/current/n1ql/n1ql-language-reference/aggregatefun.html#count), [MIN()](https://docs.couchbase.com/server/current/n1ql/n1ql-language-reference/aggregatefun.html#min), [MAX()](https://docs.couchbase.com/server/current/n1ql/n1ql-language-reference/aggregatefun.html#max), or [SUM()](https://docs.couchbase.com/server/current/n1ql/n1ql-language-reference/aggregatefun.html#sum).

That's why the reftext for this function is set to `ARRAY_APPEND()`.

Can you try include the empty brackets in the function name here and throughout, so we can see if that still looks too crowded in the right-hand navigation?

modules/n1ql/pages/n1ql-language-reference/arrayfun.adoc
```diff
@@ -47,17 +53,23 @@ SELECT ARRAY_AGG( [ "abc", 1, NULL ] ) AS array_aggregate;
...
 
 === Description
 This function takes an array [.var]`expr` and one or more [.var]`val` arguments to return a new array with the specified [.var]`val` argument(s) appended.
 
 It requires a minimum of two arguments and returns an error if there are fewer.
 
-=== Arguments
+=== Syntax
+[subs=normal]
```
simon-dew: You don't need `[subs=normal]` if there is no inline formatting in the literal block.

I'm still a bit conflicted about having two separate styles for syntax descriptions in the SQL++ docs, i.e. the EBNF style for keywords, and whatever this is for functions. I'll think about that some more.

modules/n1ql/pages/n1ql-language-reference/arrayfun.adoc
```diff
@@ -239,24 +269,30 @@ AS combined_array;
...
 expr:: [Required] The array to be searched.
 
 val:: [Required] The value that is being searched for.
 
 === Return Values
-If either of the input argument types are `MISSING`, then it returns `MISSING`.
+If either of the input arguments is `MISSING`, it returns `MISSING`.
 
-If either of the input argument types are `NULL`, then it returns `NULL`.
+If either of the input argument types is `NULL`, it returns `NULL`.
```
simon-dew: Just wondering why you've left "types" here but removed it above?


```suggestion
If either of the input arguments is `NULL`, it returns `NULL`.
```

modules/n1ql/pages/n1ql-language-reference/arrayfun.adoc
```diff
@@ -630,14 +707,20 @@ AS array_intersection;
 ====
 
 [[fn-array-length,ARRAY_LENGTH()]]
-== ARRAY_LENGTH([.var]`expr`)
+== ARRAY_LENGTH
 
-_Equivalent_: xref:n1ql-language-reference/metafun.adoc#len[LEN()]
+This function has a synonym xref:n1ql-language-reference/metafun.adoc#len[LEN()]
```
simon-dew: If we're getting rid of the **Description** heading, then in cases where there is a synonym we should probably put the synonym paragraph after the description.

modules/n1ql/pages/n1ql-language-reference/bitwisefun.adoc
```diff
@@ -967,6 +1023,6 @@ This returns 10 (1010 in binary) because 3 XOR 6 equals 5 (0101 in binary), and
 ====
 
 [#IsBitSET]
-== ISBITSET → see BITTEST
+== ISBITSET
```
simon-dew: See [Vector Functions › Aliases](https://docs.couchbase.com/server/current/n1ql/n1ql-language-reference/vectorfun.html#aliases) for another way of dealing with function aliases. It removes all the aliases from the right-hand nav which leads to neater navigation. Might be worth adopting consistently throughout?

simon-dew (dismissed):
It's great to see you overhauling the functions documentation, thank you. Some comments which may need further thought or discussion

modules/n1ql/pages/n1ql-language-reference/arrayfun.adoc
rakhi-prathap: Yes, sounds good. 
It's done!

modules/n1ql/pages/n1ql-language-reference/arrayfun.adoc
rakhi-prathap: Done.

modules/n1ql/pages/n1ql-language-reference/arrayfun.adoc
rakhi-prathap: Done. The preview site now reflects this change.

modules/n1ql/pages/n1ql-language-reference/arrayfun.adoc
rakhi-prathap: Removed `[subs=normal]`.

modules/n1ql/pages/n1ql-language-reference/arrayfun.adoc
rakhi-prathap: Done.

modules/n1ql/pages/n1ql-language-reference/bitwisefun.adoc
rakhi-prathap: Done.

simon-dew (approved):
LGTM 👍

PR #565, opened 2026-01-23
[DOC-13222]: Feedback on Reserved Words | Couchbase Docs

RayOffiah:
Changed the table to a list of entries with a heading for each section.

Folk can now navigate from the right-hand menu.

Preview:

https://preview.docs-test.couchbase.com/docs-devex-DOC-13222/release/8.0/Navigation-menu/server/current/n1ql/n1ql-language-reference/reservedwords.html

### DIFF
--- a/modules/n1ql/pages/n1ql-language-reference/reservedwords.adoc
+++ b/modules/n1ql/pages/n1ql-language-reference/reservedwords.adoc
@@ -2,12 +2,13 @@
 :description: {sqlpp} defines an extensive list of keywords that are reserved words. \
 You cannot use these keywords as identifiers unless you escape them.
 :page-topic-type: reference
+:page-toclevels: 3
 
 [abstract]
 {description}
-All of the {sqlpp} keywords are case insensitive.
+All of the {sqlpp} keywords are case-insensitive.
 
-Some of the keywords are not currently implemented but are reserved for future use.
+Some keywords are not currently implemented but are reserved for future use.
 
 == Using Reserved Words as Identifiers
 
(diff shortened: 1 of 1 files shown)


modules/n1ql/pages/n1ql-language-reference/reservedwords.adoc
```diff
@@ -33,282 +34,291 @@ CREATE INDEX myindex ON default(`index`) USING GSI;
...
-| xref:n1ql:n1ql-language-reference/begin-transaction.adoc[WORK]
-| XOR
-|
-|
-|
-|
-
-|===
+[cols=1]
+=== _
```
simon-dew: `-_` looks a bit wierd in the right-hand nav, could we change it the title to Symbols?


```suggestion
=== Symbols
```

simon-dew (changes requested):
I think it looks loads better now. We could now put the keywords in a columnar layout underneath each letter heading, if we can work out an automatic way to do it? Referring to @osfameron for ideas

modules/n1ql/pages/n1ql-language-reference/reservedwords.adoc
```diff
@@ -33,282 +34,291 @@ CREATE INDEX myindex ON default(`index`) USING GSI;
...
+=== X
+XOR
+
+
+
+
+
+
 
 NOTE: The word `AI` is not a reserved word, even though you can use it in combination with the `USING` keyword as part of the xref:n1ql-language-reference/using-ai.adoc[USING AI] statement.
```
simon-dew: Also maybe should move this note to the end of the **A** section. People might miss it down here

RayOffiah:
Hi there. 👋🏾

Right, I've used a slab of CSS to implement the change. (Thanks for the suggestion, Hakim).

Now there are a few of choices here:

We could add the CSS to `docs-ui`.

We could implement an extension in AsciiDoc. Something like:

```asciidoc
[columns, width="70%", gap="10px", cols="4"]
--
Text to flow in columns
--
```

Or, my favourite suggestion: leave it as a block on the one page that uses it. I don't think that this is going to come up so often that it's worth doing one of the first two.

modules/n1ql/pages/n1ql-language-reference/reservedwords.adoc
```diff
@@ -31,284 +32,399 @@ CREATE INDEX myindex ON default(`index`) USING GSI;
...
-You do not have to escape the word `AI` inside backticks when using it by itself as an identifier.
\ No newline at end of file
+You do not have to escape the word `AI` inside backticks when using it by itself as an identifier.
+
+++++
+<style>
+
+.openblock.columns .content .paragraph {
+  column-count: 4;
+  width: 70% ;
```
osfameron: The table currently on https://docs.couchbase.com/server/current/n1ql/n1ql-language-reference/reservedwords.html uses the full width, and I think that looks better tbh.

Did you experiment with `column-count: auto` at all? (It might require an explicit width to be set on the content divs)

osfameron (approved):
I think the 🔤 headings work nicely, and this is overall an improvement as requested in the ticket, so I'll approve, but a few caveats/comments follow:

* Happy to defer to Simon if he has further reservations
* I'm not sure that the extra CSS is that much of an improvement over for example separate tables with `[cols=4*]`
* It might be worth trying column-count:auto, which would at least give us some responsiveness to the layout. If that's useful generally, we could incorporate this into the UI bundle.
* I feel like having to manually add the A-Z headers is a bit of a faff. Certainly, this exists over a bunch of Mobile pages, where it's a nightmare to maintain. If we have a use-case for **Glossaries** and other alphabetically tagged catalogues like this, then we could usefully specify it, and extract a general UI element. (Not urgent, but definitely food for thought.)
* What's with the branch name `DOC-13222/release/8.0/Navigation-menu` ? Is that a new thing in the server space.  I'd generally advise against `/` in branch names as the folder structure can cause some Git gotchas.  Our use of e.g. `release/8.0` is too embedded so I've not suggested moving away from it, but I'd prefer not to start using the pattern elsewhere unless we really need to!

modules/n1ql/pages/n1ql-language-reference/reservedwords.adoc
simon-dew: Agree, let's try full-width

modules/n1ql/pages/n1ql-language-reference/reservedwords.adoc
```diff
@@ -31,284 +32,399 @@ CREATE INDEX myindex ON default(`index`) USING GSI;
...
+}
+
+</style>
+++++
+
+=== Symbols
+[.columns]
+--
+[%hardbreaks]
+_INDEX_CONDITION
```
simon-dew, line 55: Is there any way to prevent words from breaking across columns? Or size the columns automatically to fit the content?

Currently, on my 1680×1050 display, the word `_INDEX_CONDITION` breaks across two columns, i.e. `_INDEX_CONDITI   ON`

simon-dew (dismissed):
This is looking loads better, thanks. Just one quibble about word breaks across columns, and I agree with Hakim about the overall width of the four-column sections.

modules/n1ql/pages/n1ql-language-reference/reservedwords.adoc
simon-dew: I don't know about `column-count: auto` though — I think four is the right number of columns. But I have a quibble about word breaking across columns, or possibly automatic column widths.

RayOffiah: > I'm not sure that the extra CSS is that much of an improvement over for example separate tables with [cols=4*]

Good point, but using columns is less work (I think). The original table arranged the keywords in alphabetical order, flowing left to right. To keep the table structure would mean rearranging all the keywords so they flowed down in columns.

Using columns, all I had to do was split them into groups and remove the `|` character.

> I feel like having to manually add the A-Z headers is a bit of a faff. Certainly, this exists over a bunch of Mobile pages, where it's a nightmare to maintain. If we have a use-case for Glossaries and other alphabetically tagged catalogues like this, then we could usefully specify it, and extract a general UI element. (Not urgent, but definitely good for thought.)

Yes, definitely worth thinking about. 

> What's with the branch name DOC-13222/release/8.0/Navigation-menu ? Is that a new thing in the server space. I'd generally advise against / in branch names as the folder structure can cause some Git gotchas. Our use of e.g. release/8.0 is too embedded so I've not suggested moving away from it, but I'd prefer not to start using the pattern elsewhere unless we really need to!

Point taken! Just because it renders Git structures as folders in IntelliJ, doesn't make it a good idea. I'll go back to just separating the branch name with dashes or something.

And I'll see what it looks like with an `auto` column count.

RayOffiah: MMmmm. 🤔

`auto` just sets it to a single column, which makes the page rather long.

osfameron: Ah well, I think it'd require setting a width on the divs themselves.
Don't worry about it, especially as Simon approves 4 as the Right Number 👍

simon-dew (approved):
Looks much better and hopefully more navigable, thanks @RayOffiah . The column breaking issue still occurs on screens narrower than about 530 px, but all kinds of other much more important aspects of the UI become unusable when the screen is that narrow, so I'm happy to let that slide for now 😁

PR #575, opened 2026-01-29
[DOC-13923] Clarify Behaviour When Field Names Contain Periods

sarahlwelton:
Ticket came in asking if we could clarify for users what happens when their field names contain periods and they try to use those with mappings in the Search Service. 

Decided best course of action was a new page describing the mapping types in more details, with the impact periods have on each type. 

Updated links and removed redundant info to support that.

Preview URL:
https://preview.docs-test.couchbase.com/docs-devex-DOC-13923-field-names-with-periods/server/current/search/about-mappings.html


modules/search/pages/about-mappings.adoc
```diff
@@ -0,0 +1,142 @@
...
+
+[#collections]
+== Collection Mappings
+
+The mapping for a collection is also called a type mapping.
+
+A type mapping includes or excludes specific documents in a collection from an index.
+A document must match the collection and document type set by a type mapping to be included in a Search index.
+
+You can control the type of a document by xref:search:set-type-identifier.adoc[setting a type identifier].
```
abhinavdangeti: Worth stating this is `optional` perhaps.

modules/search/pages/about-mappings.adoc
```diff
@@ -0,0 +1,142 @@
...
+
+Type mappings can be <<static,>> or <<dynamic,>>.
+Add <<objects,>> and <<fields,>> to make a collection type mapping a static mapping.
+
+For more information about how to create a type mapping in the standard editor, see xref:search:create-type-mapping.adoc[].
+For more information about how to configure settings for mappings and type mappings in the Quick Editor, see xref:quick-index-field-options.adoc[].
+
+=== Collection Names
+
+You cannot use a collection name that contains a period (`.`) in a Search index type mapping.
```
abhinavdangeti: KV forbids this already - so it shouldn't be a problem downstream.

modules/search/pages/about-mappings.adoc
```diff
@@ -0,0 +1,142 @@
...
+
+For more information about how to create a type mapping in the standard editor, see xref:search:create-type-mapping.adoc[].
+For more information about how to configure settings for mappings and type mappings in the Quick Editor, see xref:quick-index-field-options.adoc[].
+
+=== Collection Names
+
+You cannot use a collection name that contains a period (`.`) in a Search index type mapping. 
+
+The Search Service uses periods (`.`) to separate collection names from xref:search:customize-index.adoc#type-identifiers[type identifiers]. 
+If you used a collection name that contained a period, the Search Service would interpret anything after the period as a filter value on your type mapping.
```
abhinavdangeti: Ditto.

modules/search/pages/about-mappings.adoc
```diff
@@ -0,0 +1,142 @@
...
+
+The Search Service uses periods to maintain object and child field relationships after flattening. 
+For example, the `country` and `region` fields inside the `origin` object would be represented as `origin.country` and `origin.region` inside your Search index: 
+
+[source,json]
+----
+include::example$about-mappings.jsonc[lines=1..8]
+----
+
+If you tried to create an object mapping for another object called `origin.country`, the Search Service would return results from that mapping alongside the nested `country` field in the `origin` object.
```
abhinavdangeti: Not sure I quite follow this line.

- What ends up happening in this situation is we'd have same field index for both the fields with name - `origin.country` and `country` nested within `origin` field mapping.
- During evaluation of an analytic query, we could even fail to determine the right analyzer for `origin.country` if `country` were not created within `origin` - leading to indeterministic behavior.

abhinavdangeti (commented):
@sarahlwelton I've added a few comments, let me also ping @capemox here (who set up the DOC ticket) so he can share his feedback on this as well.

modules/search/pages/about-mappings.adoc
sarahlwelton: Good to know - I'm going to tweak this to specify the "collection type mapping name", instead - because I still think it's worthwhile to explain how this mechanism works.

modules/search/pages/about-mappings.adoc
sarahlwelton: Have taken another stab at explaining this - this breakdown was helpful :) The ticket was a little light on details on what the problematic mechanism actually was, so I admit I was somewhat guessing - lol.

abhinavdangeti (approved):

modules/search/pages/about-mappings.adoc
```diff
@@ -0,0 +1,142 @@
...
+
+For example, you could create a child field mapping for the `type`, `name`, and `category` fields underneath the collection type mapping for this document.
+You could also create a child field mapping under the object mapping for the `origin` object for the `country` and `region` fields:
+
+[source,json]
+----
+include::example$about-mappings.jsonc[lines=1..8]
+----
+
+Just like <<collection-names,>> and <<object-names,>>, to avoid undesired results in your Search queries, avoid using periods (`.`) in your child field names.
```
Rebecca-Martinez007: `<<collection-names>>` is broken here, I don't think `[#collection-names]` exists in this page.

modules/search/pages/about-mappings.adoc
```diff
@@ -0,0 +1,142 @@
...
+[#xattrs]
+== XATTRs Mappings
+
+IMPORTANT: You can only use XATTRs mappings on Couchbase Server version 7.6.2 or later.
+
+Extended Attributes (XATTRs) mappings add fields and metadata from your document's XATTRs and let you query them inside a Search index.
+
+XATTRs mappings can be <<static,static>> or <<dynamic,dynamic>>, based on whether you add <<fields,>>.
+
+Just like <<collection-names,>>, <<object-names,>>, and <<fields,>>, to avoid undesired results in your Search queries, avoid using periods (`.`) in the names of fields in your document metadata.
```
Rebecca-Martinez007: Same comment here: <<collection-names>> is broken here, I don't think [#collection-names] exists in this page.

Rebecca-Martinez007 (changes requested):
Just a broken link within the page to fix up, the rest LGTM!

modules/search/pages/about-mappings.adoc
sarahlwelton: AH I think I missed deleting that when Abhi suggested I make some changes earlier. Thanks for catching that.

Rebecca-Martinez007 (approved):
LGTM!

PR #599, opened 2026-02-18
[DOC-13245] Search Index Architecture

sarahlwelton:
We needed to add a page in the Search documentation to explain the architecture of an index. 

Went and added the same diagram to the existing Vector Search Index Architecture topic, since it applies to both. 

Preview URL:

* https://preview.docs-test.couchbase.com/docs-devex-DOC-13245-search-index-architecture/server/current/search/search-index-architecture.html

* https://preview.docs-test.couchbase.com/docs-devex-DOC-13245-search-index-architecture/server/current/vector-search/vector-search-index-architecture.html


modules/search/examples/diagrams/search-architecture.mermaid
```diff
@@ -0,0 +1,29 @@
...
+    end
+
+    G@{ shape: rect, label: "Disk Write Queue" }
+    Segment --> G
+    end
+    G --> Disk
+    subgraph Disk
+        H@{ shape: docs, label: "LSM Tree of Segments" }
+        I@{ shape: cyl, label: "Metadata Store" }
+    end
```
abhinavdangeti, line 28: Nice 👍🏼

modules/search/pages/search-index-architecture.adoc
```diff
@@ -0,0 +1,82 @@
...
+
+You can also control how these lists of terms, or tokens, are generated by using an xref:customize-index.adoc#analyzers[analyzer].
+
+The inverted index, term dictionary, or the vector index are stored in <<segments,>>, which are sent to the disk write queue and saved to disk through <<sync,synchronization with the Data Service>>. 
+
+[#sync]
+== Synchronization with Database Change Protocol (DCP) and the Data Service
+
+The Search Service uses batches to process data that comes in from xref:server:learn:clusters-and-availability/intra-cluster-replication.adoc#database-change-protocol[DCP] and the xref:server:learn:services-and-indexes:services/data-service.adoc[Data Service].
+DCP and Data Service changes are introduced gradually, based on available memory on Search Service nodes, until reindexing operations for an index are complete.
```
abhinavdangeti: Not sure if I follow what do you mean by "until reindexing operations for an index are complete" ?

modules/search/pages/search-index-architecture.adoc
```diff
@@ -0,0 +1,82 @@
...
+
+The Search Service can merge batches into a single batch before they're sent to the disk write queue, to reduce the resources required for batch processing. 
+
+The Search Service maintains index snapshots on each Search index partition.
+These snapshots contain a representation of document mutations on either a write queue, or in storage.
+
+If the Search Service loses connection to the Data Service, the Search Service compares its rollback sequence numbers in its snapshots with the Data Service when the connection is reestablished.
```
abhinavdangeti: No, we should _not_ talk about rollback here already.

If the search service loses connection to the data service, it'll re-initiate a connection request from last update (seqno) it persisted. If the data service for whatever reason asks search to rollback to an earlier point - this is when the search service recovers seqnos from earlier snapshots (rollback seqnos) to setup stream requests.

modules/search/pages/search-index-architecture.adoc
```diff
@@ -0,0 +1,82 @@
...
+If the index snapshots on the Search Service are too far ahead, the Search Service performs a full rollback to get back in sync with the Data Service.
```
abhinavdangeti: too far ahead compared to the data service*

modules/search/pages/search-index-architecture.adoc
```diff
@@ -0,0 +1,82 @@
...
+The Search Service maintains index snapshots on each Search index partition.
+These snapshots contain a representation of document mutations on either a write queue, or in storage.
+
+If the Search Service loses connection to the Data Service, the Search Service compares its rollback sequence numbers in its snapshots with the Data Service when the connection is reestablished.
+If the index snapshots on the Search Service are too far ahead, the Search Service performs a full rollback to get back in sync with the Data Service. 
+
+[#segments]
+== Search Index Segments
+
+Search indexes are built with index snapshots, which contain data segments.
```
abhinavdangeti: Index snapshots are more of an abstraction - so saying built with index snapshots is not right .. I'll let you rephrase this better.

modules/search/pages/search-index-architecture.adoc
```diff
@@ -0,0 +1,82 @@
...
+
+If the Search Service loses connection to the Data Service, the Search Service compares its rollback sequence numbers in its snapshots with the Data Service when the connection is reestablished.
+If the index snapshots on the Search Service are too far ahead, the Search Service performs a full rollback to get back in sync with the Data Service. 
+
+[#segments]
+== Search Index Segments
+
+Search indexes are built with index snapshots, which contain data segments. 
+
+All Search indexes contain a root index snapshot.
```
abhinavdangeti: A root index snapshot reflects the latest snapshot of the index - collection of segments that hold only the latest version of all data available .. again an abstraction.

modules/search/pages/search-index-architecture.adoc
sarahlwelton: Might have been a holdout from some other material... I think the sentence makes sense without that part, so I'll just drop it.

modules/search/pages/search-index-architecture.adoc
sarahlwelton: Going to try and add a new section that explains what happens when connection drops, instead. Do you think there's more that needs to be said about data sync and DCP otherwise in this section?

modules/search/pages/search-index-architecture.adoc
sarahlwelton: Maybe we'll just say "built with data segments."

abhinavdangeti (approved):

modules/search/examples/diagrams/search-architecture.mermaid
```diff
@@ -0,0 +1,29 @@
+[mermaid]
+----
+flowchart LR
+    subgraph Memory 
+    A@{ shape: docs, label: "JSON Documents" }
+    B@{ shape: rect, label: "Text Analysis" }
+    C@{ shape: lean-r, label: "Token Stream" }
+    E@{ shape: rect, label: "FAISS" }
+    
+    A --> B
```
RayOffiah, line 10: This is probably just a hangover from my developer days, but I think it's better to name items in a way that tells the reader what the item is for. 
For example, rather than `A`, I probably would have called it `JSONDocuments`

RayOffiah (approved):
Minor point about naming, but otherwise LGTM 👍🏾

PR #620, opened 2026-03-27
Removing alias which was preventing the build from … building.

RayOffiah:
Removed alias which was crashing the build.


osfameron (changes requested):
as discussed, probably not needed - covered in docs-server release/8.0 so that change needs to be rebased into your working branch.

PR #644, opened 2026-05-05
[DOC-10029] Update valid characters for GSI index names

rakhi-prathap:
Jira: DOC-10029

Updates the `CREATE INDEX` page to correct the valid characters listed for GSI index names.


rakhi-prathap:
Merging without review since this an is extremely small change.

PR #651, opened 2026-05-21
[DOC-3414] Fix CURL function example

rakhi-prathap:
Jira: DOC-3414

Changes: 

- Removed the restriction requiring CURL() arguments to be static values.
- Fixed a broken link that was preventing an example from rendering correctly. 

Docs preview: [CURL()](https://preview.docs-test.couchbase.com/docs-devex-DOC-3414-fix-curl-function-example/server/current/n1ql/n1ql-language-reference/curl.html#from-the-query-settings-rest-api)


modules/n1ql/pages/n1ql-language-reference/curl.adoc
```diff
@@ -297,19 +297,6 @@ The redirection of URL is not allowed.
 This means that files on the local file system cannot be accessed.
 * The amount of memory used for the CURL result is controlled using the `result-cap` option.
 The default is 20MB.
-* Any values passed to the arguments of `CURL()` must be static values.
```
rakhi-prathap, line 300: Removing this point along with the example. 

CURL() supports document field references, SQL++ expressions, and dynamic parameters. Locally tested and verified.

modules/n1ql/pages/n1ql-language-reference/curl.adoc
simon-dew, line 1: Do a Vale check: in particular change the use of "we" in the examples and also there's a "Note that"

simon-dew (changes requested):
Looks good apart from style and tone fixes

modules/n1ql/pages/n1ql-language-reference/curl.adoc
rakhi-prathap: Done 👍

PR #674, opened 2026-07-29
DOC-14139: Add ability to perform fts query from eventing

Pallavi-Janardhan:
Included information for FTS SearchQuery for Eventing DOC-14139

[Docs preview](https://preview.docs-test.couchbase.com/docs-devex-DOC-14139_Perform_FTS_Query_from_Eventing/server/current/eventing/eventing-language-constructs.html#search_call)

### DIFF
--- a/modules/eventing/pages/eventing-language-constructs.adoc
+++ b/modules/eventing/pages/eventing-language-constructs.adoc
@@ -413,6 +413,58 @@ function OnUpdate(doc, meta) {
 
 For more information about {sqlpp} Analytics, see xref:server:analytics:1_intro.adoc[].
 
+
+[#search_call]
+=== `couchbase.{zwsp}searchQuery({wj})`
+
+ifeval::['{page-component-version}' == '8.1'] 
+_(Introduced in Couchbase Server 8.1)_
+endif::[]
+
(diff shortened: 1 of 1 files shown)


modules/eventing/pages/eventing-language-constructs.adoc
```diff
@@ -413,6 +413,58 @@ function OnUpdate(doc, meta) {
...
+    // 2. Execute the search query
+    var it = couchbase.searchQuery("travel-sample._default.travel_sample_test", myQuery, {"limit": 10});
+
+    // 3. Iterate through the search results
+    for (let row of it) {
+        count++;
+        log("Match found: ", row);
+    }
+    
+    if (count > 0) {
```
AnkitPrabhu, line 459: we don't need this part and can remove count variable. That was only in testing script to match count matches the limit or not. I guess this is not needed in the example script

Pallavi-Janardhan: The original example used {"limit": 10} as the options parameter, but all the calls in the official test script (search.js) use {"size": 5} — none of them use limit. That's bit inconsistent! Hence I switched it to "size" above to match the test script, and deleted the count. Can one of you please confirm which key the searchQuery() options actually accepts? Is the new modification ok/fine?

AnkitPrabhu (approved):

saimirra-ramesh (approved):

